PR #12, opened 2016-04-10
Sample: Use create instead of getInstance

ivanph:
getInstance was renamed to create in 7ca4e0312ae898e369ba1e67bd9ec9245fb2a4d9
but wasn't changed here


uKL:
Thanks, good point!

PR #19, opened 2016-05-09
Prevent stopLeScan from being called before startLeScan.

mzgreen:
It is a proposed solution to fix https://github.com/Polidea/RxAndroidBle/issues/17

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
@@ -39,12 +39,10 @@ public class RxBleRadioOperationScan extends RxBleRadioOperation<RxBleInternalSc
 
             try {
 
-                if (!getSubscriber().isUnsubscribed()) {
-                    isScanStarted.set(rxBleAdapterWrapper.startLeScan(leScanCallback));
+                isScanStarted.set(rxBleAdapterWrapper.startLeScan(leScanCallback));
 
-                    if (!isScanStarted.get()) {
-                        onError(new BleScanException(BleScanException.BLUETOOTH_CANNOT_START));
-                    }
+                if (!isScanStarted.get()) {
+                    onError(new BleScanException(BleScanException.BLUETOOTH_CANNOT_START));
                 }
             } finally {
                 releaseRadio();
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
@@ -16,14 +16,13 @@ public class RxBleRadioOperationScan extends RxBleRadioOperation<RxBleInternalSc
     private final UUID[] filterServiceUUIDs;
     private final RxBleAdapterWrapper rxBleAdapterWrapper;
     private final UUIDUtil uuidUtil;
-    private final Object scanLock = new Object();
-    private boolean isScanStarted;
     private final BluetoothAdapter.LeScanCallback leScanCallback = (device, rssi, scanRecord) -> {
 
         if (!hasDefinedFilter() || hasDefinedFilter() && containsDesiredServiceIds(scanRecord)) {
             onNext(new RxBleInternalScanResult(device, rssi, scanRecord));
         }
     };
+    private boolean isScanStarted;
 
     public RxBleRadioOperationScan(UUID[] filterServiceUUIDs, RxBleAdapterWrapper rxBleAdapterWrapper, UUIDUtil uuidUtil) {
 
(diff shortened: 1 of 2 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
```diff
@@ -16,13 +16,13 @@
...
     private final RxBleAdapterWrapper rxBleAdapterWrapper;
     private final UUIDUtil uuidUtil;
-
     private final BluetoothAdapter.LeScanCallback leScanCallback = (device, rssi, scanRecord) -> {
 
         if (!hasDefinedFilter() || hasDefinedFilter() && containsDesiredServiceIds(scanRecord)) {
             onNext(new RxBleInternalScanResult(device, rssi, scanRecord));
         }
     };
+    private boolean startLeScanStatus;
```
uKL: What about `isScanStarted`?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
```diff
@@ -46,8 +46,11 @@ public void run() {
     }
 
     public void stop() {
+
         // TODO: [PU] 29.01.2016 https://code.google.com/p/android/issues/detail?id=160503
-        rxBleAdapterWrapper.stopLeScan(leScanCallback);
+        if (startLeScanStatus) {
```
uKL: Stopping and starting should be synchronized (on status), plus startLeScanStatus should be set to false after the scan is stopped.

mzgreen: Good point. What do you think about making status `volatile` instead of using a monitor?

dariuszseweryn: What would you want to achieve by using `volatile` here?

mzgreen: It would ensure that status value is the same across different threads. Also it's more lightweight because we would avoid context switching.

dariuszseweryn: Still there is room for a race condition. When the radio will start executing the scan operation but will not yet call `run()` and the operation would be unsubscribed then.
We could check if the operation wasn't unsubscribed right before calling `startLeScan()` but it would still need some manual synchronisation.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/RxBleClientTest.groovy
```diff
@@ -315,4 +313,31 @@ class RxBleClientTest extends Specification {
         mock.getAddress() >> scanData['deviceMac']
         bleAdapterWrapperSpy.addScanResult(mock, scanData['rssi'], scanData['scanRecord'])
     }
+
+    def "should not call stopLeScan if startLeScan wasn't called before"() {
```
dariuszseweryn: We should have a test that is currently causing us problems. Something like `def "should not startLeScan() if stop() was called before"`. Maybe it will be even easier.

mzgreen: But it's something different I guess. Calling `stopLeScan` and then `startLeScan` is not bad I guess. The bad thing is calling `stopLeScan` after a call `startLeScan` but before `startLeScan` returns.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
mzgreen: I think you're right but such window is so small that it's almost impossible to run into such edge case. I think that `volatile` or a monitor will be enough. What about you @uKL ?

mzgreen:
Ok I've added monitors and used `AtomicBoolean` because `synchronized` should be used with final fields. Scan is not a thing that will be called really frequently (like in a loop) so I think that performance is not such a big deal here.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationScan.java
```diff
@@ -34,20 +35,31 @@ public RxBleRadioOperationScan(UUID[] filterServiceUUIDs, RxBleAdapterWrapper rx
...
-            boolean startLeScanStatus = rxBleAdapterWrapper.startLeScan(leScanCallback);
+        synchronized (isScanStarted) {
+
+            try {
 
-            if (!startLeScanStatus) {
-                onError(new BleScanException(BleScanException.BLUETOOTH_CANNOT_START));
+                isScanStarted.set(rxBleAdapterWrapper.startLeScan(leScanCallback));
+
+                if (!isScanStarted.get()) {
```
uKL: If you are synchronizing whole block then atomicboolean seems to be not required anymore.

mzgreen: Yup but `synchronizer` requires object (it doesn't work with primitive types). And lint suggests that it should be a final object so a `Boolean` can't be used too.

uKL: You can go with private final Object lock = new Object[0]

mzgreen: I wanted to reuse a `status` object. But introducing additional one is an option too.

PR #62, opened 2016-08-24
Fixed https://github.com/Polidea/RxAndroidBle/issues/60

mzgreen:
Connection was emitted incorrectly in `establishConnection` method. Observable was calling `onComplete` after emission which then caused that `doOnUnsubscribe`. This is the reason that `DISCONNECTED` state was emitted every time, and it should only be emitted after explicit unsubscription from connection observable.

### DIFF
--- a/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleDeviceMock.java
+++ b/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleDeviceMock.java
@@ -58,7 +58,7 @@ class RxBleDeviceMock implements RxBleDevice {
     public Observable<RxBleConnection> establishConnection(Context context, boolean autoConnect) {
         return Observable.defer(() -> {
             if (isConnected.compareAndSet(false, true)) {
-                return Observable.just(rxBleConnection)
+                return emitConnectionWithoutCompleting()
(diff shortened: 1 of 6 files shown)


mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleDeviceMock.java
```diff
@@ -76,6 +76,12 @@ public void addAdvertisedUUID(UUID advertisedUUID) {
     }
 
     @Override
+    public BluetoothDevice getBluetoothDevice() {
```
dariuszseweryn: is this change of places needed?

mzgreen: No, just autoformatter. Do you want me to revert it to the previous state?

dariuszseweryn: Yes, for pull requests these reorderings only make it worse to read. For refactorings that are not changing code but only reordering - you can push directly.

mzgreen:
Done, also I've rebased to current master in order to avoid checkstyle fail due to an unused import in other part of the library.

dariuszseweryn:
Rebased and merged into master ( https://github.com/Polidea/RxAndroidBle/commit/f1a67da064d7874a9ae3cd67240792bc28094026 )

PR #80, opened 2016-10-14
On the fly mock device discovery

nightscape:
I have a use-case where I need to simulate devices being discovered only after a certain time.
This PR adds a `RxBleClientMock#simulateDeviceDiscovery` method that can be used for this.
I'm relatively new to RxJava, but I'm hoping that the use of a `ReplaySubject` gives the same `scanBleDevices` behaviour as before. If there are better ways to do this, let me know :slightly_smiling_face: 
I also needed to make  `RxBleDeviceMock` and `RxBleConnectionMock` public in order to be able to create instances of those classes outside the `addDevice` method.

### DIFF
--- a/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleClientMock.java
+++ b/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleClientMock.java
@@ -30,15 +30,20 @@ public class RxBleClientMock extends RxBleClient {
 
     public static class Builder {
 
-        private Map<String, RxBleDevice> discoverableDevices;
+        private ReplaySubject<RxBleDeviceMock> discoverableDevicesSubject;
         private Set<RxBleDevice> bondedDevices;
 
         /**
          * Build a new {@link RxBleClientMock}.
          */
         public Builder() {
-            this.discoverableDevices = new HashMap<>();
+            this.discoverableDevicesSubject = ReplaySubject.create();
             this.bondedDevices = new HashSet<>();
+        }
+
+        public Builder setDeviceDiscoveryObservable(@NonNull Observable<RxBleDeviceMock> discoverableDevicesObservable) {
+            discoverableDevicesObservable.subscribe(this.discoverableDevicesSubject);
+            return this;
         }
 
         /**
@@ -47,7 +52,7 @@ public class RxBleClientMock extends RxBleClient {
          * @param rxBleDevice device that the mocked client should contain. Use {@link DeviceBuilder} to create them.
          */
         public Builder addDevice(@NonNull RxBleDevice rxBleDevice) {
-            discoverableDevices.put(rxBleDevice.getMacAddress(), rxBleDevice);
+            this.discoverableDevicesSubject.onNext((RxBleDeviceMock) rxBleDevice);
             return this;
         }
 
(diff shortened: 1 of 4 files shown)


mzgreen:
Hi, thanks for your contribution and sorry for the delay! In general it looks good, the only thing that I don't like is exposing `RxBleDeviceMock` and `RxBleConnectionMock` classes. Is it really necessary? What is your use case?

uKL:
Hey @nightscape, would you be so kind to do some follow up to the question?

nightscape:
Our use case is that we start a device discovery and wait for a first device to be found. If after this first device a second device is found, we show a dialog to let the user choose between the two. If no other device is found, we wait for 5 seconds and then continue with the first device.
A further test would then check that the connection to the chosen device was established correctly and thet we are reading from that device.

I could hide the `RxBleDeviceMock` and `RxBleConnectionMock` behind interfaces that extend `RxBleDevice` and `RxBleConnection` and expose methods for stubbing data retrieval etc, but I guess that would only slightly improve things.

nightscape:
@mzgreen @uKL I just rebased my changes on the latest master branch.
Are there any open questions or anything I can do to move this forward?

mzgreen:
Looks good to me 👍

uKL:
Looks good to me either.

mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleClientMock.java
```diff
@@ -271,18 +277,25 @@ public RxBleDevice getBleDevice(@NonNull String macAddress) {
         return createScanOperation(filterServiceUUIDs);
     }
 
+    public void simulateDeviceDiscovery(@NonNull RxBleDeviceMock deviceMock) {
```
uKL: One more thing. The method name suggests that this is one time operation, I mean that it will emit provided device and will not affect future scans. By use of a ReplaySubject each device is stored and then magically replied during following scan operations.

Maybe we could have to separate configurations. Static - configured when the mock is instantiated (stored in discoverableDevices) and dynamic, on demand. You may use a PublishSubject instead and do following in line #291

`
discoveredDevicesSubject
.startWith(Observable.from(discoverableDevices))
`

dariuszseweryn: As we talked. I am also thinking if it would not be good to add method `simulateScanError(@NonNull Throwable throwable)`. But it is a different topic.

nightscape: Another maybe even nicer option would be to add a constructor that directly takes an `Observable` that gets assigned to the `discoveredDevicesSubject` field. That would give users complete freedom on when which devices appear and disappear (and maybe even the scan error stuff?).
I just tried to stick close to the current implementation in order to increase the chances to get this in ;)

uKL: That's a good idea :) What I'd do is to provide an observable to the mock. In the builder there would be two options:

- public Builder addDevice(@NonNull RxBleDevice rxBleDevice)
- public Builder setDeviceDiscoveryObservable(@NonNull Observable<RxBleDevice> rxBleDevice)

Then the builder with pass merged data feed to the mock. This would be really flexible and easy to use.

dariuszseweryn: /agree - would be a great, and simple, addition

uKL:
I will merge this pull request and then apply some small changes. Thank you for commitment! :)

nightscape:
Hi @uKL, thanks for your support and merging!
One note about your  comment on `PublishSubject` vs `ReplaySubject`:
Currently the method `createScanOperation` and `getBleDevice` rely on the fact that the discovered devices are memoized and can be replayed to new observers (which I assume the `filter` and `first` methods create). Probably there is a nicer option to do this, I'm a relative Rx noob :wink:

uKL:
Yes, I noticed that. We will address that issue :)

uKL:
Dear Contributor,

PR #89, opened 2016-11-01
add unpair functionality

ghost:
Added the ability to unpaid a device programmatically.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleDeviceImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleDeviceImpl.java
@@ -6,7 +6,9 @@ import android.content.Context;
 import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDevice;
 import com.polidea.rxandroidble.exceptions.BleAlreadyConnectedException;
+import com.polidea.rxandroidble.exceptions.BleDeviceUnpairException;
 
+import java.lang.reflect.Method;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import rx.Observable;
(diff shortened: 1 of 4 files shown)


aldoborrero:
Hi @julien-fueled does this make sense to put this method here? I mean, when you "unpair" is for unboding a device. A device can be "unbonded" if the programmer has explicitly called "bond".

The problem itself is that bonding process is dependent on the device itself, for example when bonding with mine I didn't have to request any pin info while others BLE devices will ask you for a pin.

Maybe for managing these use cases we can try to create an external helper class that will aid in the process of bonding/unbonding and leave this class only for connecting to the device.

I have here [a basic gist](https://gist.github.com/aldoborrero/3dee9dfd2ec6c808ac56a63ae5b1de02) which implements this as a helper class (obviously it can be improved).

What do you think @dariuszseweryn about this?

dariuszseweryn:
Hello,
First of all I would like to thank you for your effort.
I have some doubts about this PR unfortunately:
- I don't think that adding only unbond feature is feasible - there should be a complex feature for whole bonding / unbonding that would fully cover the functionality (bonding, bonding with pin, ...)
- The nomenclature in Bluetooth is to `bond` not `pair`
- The javadoc doesn't match the style of the rest of the library
- In the current implementation there is no value added by returning `Observable` - it could be a synchronous call that is really easy to wrap into an observable using `Observable.fromCallable()`
I have been fiddling with this feature for some time but wasn't able to achieve a standard behaviour among different devices. Now I am focusing on fixing bugs to stabilise the library and features will be worked on later. If you have time - feel free to work on it.
Best Regards

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleDeviceImpl.java
```diff
@@ -42,17 +44,17 @@ public RxBleDeviceImpl(BluetoothDevice bluetoothDevice, RxBleConnection.Connecto
...
-                            .doOnSubscribe(() -> connectionStateSubject.onNext(CONNECTING))
-                            .doOnNext(rxBleConnection -> connectionStateSubject.onNext(CONNECTED))
-                            .doOnUnsubscribe(() -> {
-                                connectionStateSubject.onNext(DISCONNECTED);
-                                isConnected.set(false);
-                            });
-                } else {
-                    return Observable.error(new BleAlreadyConnectedException(bluetoothDevice.getAddress()));
-                }
+            if (isConnected.compareAndSet(false, true)) {
```
piotrek1543: I don't see where `isConnected` is set to `true` - it seems to be always `false`

dariuszseweryn: At this very line it is set to `true` if the current state equals `false`

RobLewis:
Is there any progress on this? My new project is an app that connects to 3 different BLE devices. One of the devices must be "paired" to operate; it is optional on the others, apparently. 
I am still puzzling over the terminology of _pairing_ vs. _bonding_, and can't find a definitive explanation. Some sources emphasize that the two are quite different things, and yet the Android documentation treats them as though they are synonyms (see [here](https://developer.android.com/reference/android/bluetooth/BluetoothDevice.html#ACTION_BOND_STATE_CHANGED), for example). 
On the assumption that when people say "pair", they really mean "bond", there seems to be a need to add this to `RxAndroidBle`. Or can I simply call `rxBleDevice.getBluetoothDevice.createBond()`? 
Clarification appreciated.

dariuszseweryn:
This PR will not make it to the `master` branch in the current form. Since the author did not add anything since the original commit I am closing this.

>  Or can I simply call `rxBleDevice.getBluetoothDevice.createBond()`?

Since the library does not support creating a bond you can use the above method or rely on the automatic process which is kicked off by Android usually when accessing encrypted characteristics.

RobLewis:
@dariuszseweryn 

> …rely on the automatic process which is kicked off by Android usually when accessing encrypted characteristics.

Is [this](https://stackoverflow.com/questions/23674668/android-bluetooth-low-energy-characteristic-getpermissions-returns-0#23831746) what you mean? 

> it seems Android attempts to handle authentication/permission issues on a trial and error basis. In other words, the framework always attempts a basic read/write, and if it fails for authentication reasons, it automatically tries again with MITM authentication requested. 

If you have a link to more info, I'd appreciate it.

dariuszseweryn:
Yes, that is it. Unfortunately I do not have any link to provide you with.

PR #99, opened 2016-11-21
+add characteristic write delay function. In some devices, such as Xi…

nziyouren:
+add characteristic write delay function. In some devices, such as Xiaomi (HM 2A, MI 4LTE, Redmi 3S). When send long data （> 20 bytes, split about 30 packets）, the peripheral device can only receive part data, even if the write operations are queued. Add time delay of write operation can solve this issue.

### DIFF
--- a/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
+++ b/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
@@ -142,6 +142,12 @@ class RxBleConnectionMock implements RxBleConnection {
     }
 
     @Override
+    public Observable<byte[]> writeCharacteristic(@NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic,
+            @NonNull byte[] data, long delay, TimeUnit unit) {
+        return Observable.just(data).delay(delay, unit);
+    }
+
+    @Override
     public Observable<byte[]> writeCharacteristic(@NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic, @NonNull byte[] data) {
         return Observable.just(data);
     }
@@ -151,6 +157,15 @@ class RxBleConnectionMock implements RxBleConnection {
         return getCharacteristic(characteristicUuid)
                 .map(characteristic -> characteristic.setValue(data))
                 .flatMap(ignored -> Observable.just(data));
+    }
+
+    @Override
+    public Observable<byte[]> writeCharacteristic(@NonNull UUID characteristicUuid, @NonNull byte[] data, long delay,
+            TimeUnit unit) {
+        return getCharacteristic(characteristicUuid)
+                .map(characteristic -> characteristic.setValue(data))
+                .flatMap(ignored -> Observable.just(data))
+                .delay(delay, unit);
     }
 
     @Override
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -209,6 +209,19 @@ public interface RxBleConnection {
     Observable<byte[]> writeCharacteristic(@NonNull UUID characteristicUuid, @NonNull byte[] data);
 
     /**
+     * Performs GATT write operation on a characteristic with given UUID.
+     *
+     * @param characteristicUuid Requested characteristic UUID.
+     * @param delay Delay time
+     * @param unit Delay time unit
+     * @return Observable emitting characteristic value after write or an error in case of failure.
+     * @throws BleCharacteristicNotFoundException if characteristic with given UUID hasn't been found.
+     * @throws BleGattCannotStartException        if write operation couldn't be started for internal reason.
+     * @throws BleGattException                   if write operation failed
+     */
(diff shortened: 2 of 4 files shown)


dariuszseweryn:
Hello @nziyouren 
Thank you for your idea and contribution. From what I see your changes are effectively an equivalent  of `RxBleConnection.writeCharacteristic().delay(long, TimeUnit)`. 

The possible problem is that your peripheral device is not managing to consume the sent data in time. The delay will be peripheral device dependent and therefore it cannot have a one-size-fit-all solution inside the library. 

Since there is no automatic splitting of bigger chunks of data when sending to a peripheral inside the library and it is managed in the application code - the needed delay may be added there as well and it is an easier / smaller change there than in the library.

The best solution would be to have some flow-control on the peripheral to know when it is ready to receive the next chunk of data.

@uKL what do you think about this PR?
@nziyouren Thank you for your effort but currently I do not see a reason to include this change.

Best Regards

nziyouren:
@dariuszseweryn 
First of all, thank you all guys for this great library. I used it in my product. It saves me a lot of time.
Why I add this function in library, I have some reasons:
1.  When I send long data, peripheral device can only receive part data. The essential reason is that , mobile phone drops some data, doesn't send it out,  **even if GATT callback method: public void onCharacteristicWrite(BluetoothGatt gatt, BluetoothGattCharacteristic characteristic, int status) reports it writes all data successfully**. So the problem is mobile phone side, not on the peripheral. We have proved it with USB dongle and ble sniffer tool.

2. Application code and RXAndroidBle library are just like producer and consumer, application call writeCharacteristic() to produce send data actions and library consumers these write actions and execute them one by one. So I think the right solution is : control the consumer speed in this situation.
**library execute write data->wait a little time-> release semaphore->execute next write data.**
In my opinion, if add this function into library, it would save developers some time. He can set the delay time according to the device. If I'm wrong, please correct me. 

Here are part of my code in my library:

    @Override
    public void writeCharacteristic(UUID uuid, byte[] data, BleStep step){
        mRxBleConnection.writeCharacteristic(uuid,data)
                .delay(50, TimeUnit.MILLISECONDS) //add here useless
                .observeOn(AndroidSchedulers.mainThread())
                .doOnNext(bytes -> notifyInteractUpdate(mBleDevice,new BleStep("Begin step "+step.action,step.rawData,step.data)))
                .subscribe(bytes -> notifyInteractUpdate(mBleDevice,new BleStep("Step "+step.action+" complete",step.rawData,step.data))
                        ,throwable -> notifyInteractError(mBleDevice,throwable,step));

    }

    @Override
    public void writeCharacteristic(UUID uuid, byte[] data){
        mRxBleConnection.writeCharacteristic(uuid,data)
                .delay(50, TimeUnit.MILLISECONDS) //add here useless
                .observeOn(AndroidSchedulers.mainThread())
                .doOnNext(bytes -> notifyInteractUpdate(mBleDevice,new BleStep("Begin write data to UUID :"+uuid.toString())))
                .subscribe(bytes -> notifyInteractUpdate(mBleDevice,new BleStep(String.format("Step write data to UUID: %s complete",uuid.toString())))
                        ,throwable -> notifyInteractError(mBleDevice,throwable,new BleStep(String.format("Step write data to UUID: %s error",uuid.toString()))));

    }

dariuszseweryn:
I would like to see some more information about this issue somewhere (from stackoverflow or other) - from my knowledge once the `BluetoothGattCallback.onCharacteristicWrite()` is invoked the `BluetoothGatt` should be able to handle next `.writeCharacteristic()` call. I may be wrong though - never had an issue like this before.

Sending a long `byte[]` can be easily made with a programmable delay with a kind of extension function on the application side:

    Observable<byte[]> writeCharacteristic(
            RxBleConnection connection,
            UUID characteristicUuid,
            byte[] aLotOfBytes,
            long delay,
            TimeUnit timeUnit
    ) {
        return chunkify(aLotOfBytes) // splitting the byte array into smaller pieces
                .flatMap(
                        bytesChunk -> connection
                                .writeCharacteristic(characteristicUuid, bytesChunk) // sending the bytes
                                .delay(delay, timeUnit), // and waiting for a specific time after each operation
                        1 // to make the flatMap executed sequentially
                );
    }

where `chunkify(byte[])` could look like this:

    Observable<byte[]> chunkify(byte[] bytes) {
        return Observable.create(new SyncOnSubscribe<ByteBuffer, byte[]>() {

            @Override
            protected ByteBuffer generateState() {
                return ByteBuffer.wrap(bytes);
            }

            @Override
            protected ByteBuffer next(ByteBuffer state, Observer<? super byte[]> observer) {
                final int chunkLength = Math.min(state.remaining(), 20);
                if (chunkLength > 0) {
                    final byte[] bytesChunk = new byte[chunkLength];
                    state.get(bytesChunk);
                    observer.onNext(bytesChunk);
                }
                if (chunkLength == 0 || state.remaining() == 0) {
                    observer.onCompleted();
                }
                return state;
            }
        });
    }

nziyouren:
@dariuszseweryn 
Sorry for the delay, I was busy on project. By the way, I'm a new bie to rxjava. Thanks for your chunkify rxjava code. I use it in my project, it works .I agreen with you , it is more reasonable to add delay to application layer.

dariuszseweryn:
No problem. I am glad I could help. 
As for now I don't see a good reason to use this PR - closing for now. Maybe in the future I will revisit my position. Anyway - if you will write some code you think will be feasible to use - feel free to create a Pull Request.
Cheers!

PR #100, opened 2016-11-22
RxBleConnection: Enable long write and MTU change support

bmc08gt:
Change-Id: Ic726f310542b2746f334b07457960ec5df559a8a
Signed-off-by: Brandon McAnsh <bmcansh@powerley.com>

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -209,6 +209,19 @@ public interface RxBleConnection {
     Observable<byte[]> writeCharacteristic(@NonNull UUID characteristicUuid, @NonNull byte[] data);
 
     /**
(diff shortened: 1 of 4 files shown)


dariuszseweryn:
Hello @bmc08gt 
Thank you for your pull request - it is a good idea to add the MTU negotiation to the library.
It would be best to not to bind MTU changes with writing to a characteristic but rather keep it as a separate function / operation as the user may want to do this prior to writing / reading. As the library is aiming to be fully tested - this should be tested as well.
Good that you have picked up the idea - I will happily merge your changes when the changes will be done.
Best Regards

bmc08gt:
Can gladly take care of that change

dariuszseweryn:
FYI There is another PR about the MTU #102 which is a bit closer to what would fit into the library.

dariuszseweryn:
MTU negotiation PR #102 was merged today. Closing. Thank you for your effort.

PR #102, opened 2016-11-24
Added support for requestMtu

pregno:
Since API 21, is possible to set the MTU connection parameter and send packet with length up to 512 bytes.

This implementation bring the support to the RxAndroidBle library.

I've also added the related tests method and updated the sample.
If you prefer I can set up a different activity in the sample for test this functionality.

I've also added the request for the location permission in the sample project.

Tested on a Nexus 5X with Nougat

I know that there is already another PR related to the MTU, but I've implemented as a separated function.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -300,7 +300,7 @@ public interface RxBleConnection {
     /**
      * Performs GATT request MTU
      *
-     * Timeouts after specified amount of time.
+     * Timeouts after 10 seconds.
      *
      * @return Observable emitting result the MTU requested.
      * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -294,4 +294,32 @@ public interface RxBleConnection {
      * @return Observable emitting the read RSSI value
      */
     Observable<Integer> readRssi();
+
+    /**
+     * Performs GATT request MTU
+     * <p>
+     *
+     * Timeouts after specified amount of time.
+     *
+     * @return Observable emitting result the MTU requested.
+     * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
+     *                                     the MTU for internal reasons.
+     * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
+     */
+    Observable<Integer> requestMtu(int mtu);
+
+    /**
+     * Performs GATT request MTU
+     * <p>
+     *
+     * Timeouts after specified amount of time.
+     *
+     * @param timeout multiplier of TimeUnit after which the discovery will timeout in case of no return values
+     * @param timeUnit TimeUnit for the timeout
+     * @return Observable emitting result the MTU requested.
+     * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
+     *                                     the MTU for internal reasons.
+     * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
+     */
+    Observable<Integer> requestMtu(int mtu, long timeout, TimeUnit timeUnit);
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -3,7 +3,9 @@ package com.polidea.rxandroidble;
 import android.bluetooth.BluetoothGattCharacteristic;
 import android.bluetooth.BluetoothGattDescriptor;
 import android.content.Context;
+import android.os.Build;
 import android.support.annotation.NonNull;
+import android.support.annotation.RequiresApi;
 
 import com.polidea.rxandroidble.exceptions.BleCannotSetCharacteristicNotificationException;
 import com.polidea.rxandroidble.exceptions.BleCharacteristicNotFoundException;
@@ -297,7 +299,6 @@ public interface RxBleConnection {
 
     /**
      * Performs GATT request MTU
-     * <p>
      *
      * Timeouts after specified amount of time.
      *
@@ -306,20 +307,6 @@ public interface RxBleConnection {
      *                                     the MTU for internal reasons.
      * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
      */
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     Observable<Integer> requestMtu(int mtu);
-
-    /**
-     * Performs GATT request MTU
-     * <p>
-     *
-     * Timeouts after specified amount of time.
-     *
-     * @param timeout multiplier of TimeUnit after which the discovery will timeout in case of no return values
-     * @param timeUnit TimeUnit for the timeout
-     * @return Observable emitting result the MTU requested.
-     * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
-     *                                     the MTU for internal reasons.
-     * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
-     */
-    Observable<Integer> requestMtu(int mtu, long timeout, TimeUnit timeUnit);
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
@@ -12,7 +12,6 @@ import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDeviceServices;
 import com.polidea.rxandroidble.exceptions.BleCannotSetCharacteristicNotificationException;
 import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetException;
-import com.polidea.rxandroidble.exceptions.BleException;
 import com.polidea.rxandroidble.internal.RxBleRadio;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicRead;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicWrite;
@@ -69,7 +68,6 @@ public class RxBleConnectionImpl implements RxBleConnection {
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private Observable<Integer> privateRequestMtu(int mtu, long timeout, TimeUnit timeUnit) {
         final Observable<Integer> newObservable;
-        // performing actual discovery
         newObservable = rxBleRadio
                 .queue(new RxBleRadioOperationMtuRequest(
                         mtu,
@@ -135,8 +133,7 @@ public class RxBleConnectionImpl implements RxBleConnection {
     }
 
     @Override
-    public Observable<Observable<byte[]>> setupNotification(
-            @NonNull BluetoothGattCharacteristic characteristic) {
+    public Observable<Observable<byte[]>> setupNotification(@NonNull BluetoothGattCharacteristic characteristic) {
         return setupServerInitiatedCharacteristicRead(characteristic, false);
     }
 
@@ -146,8 +143,7 @@ public class RxBleConnectionImpl implements RxBleConnection {
     }
 
     @Override
-    public Observable<Observable<byte[]>> setupIndication(
-            @NonNull BluetoothGattCharacteristic characteristic) {
+    public Observable<Observable<byte[]>> setupIndication(@NonNull BluetoothGattCharacteristic characteristic) {
         return setupServerInitiatedCharacteristicRead(characteristic, true);
     }
 
@@ -255,8 +251,7 @@ public class RxBleConnectionImpl implements RxBleConnection {
     }
 
     @Override
-    public Observable<byte[]> readCharacteristic(
-            @NonNull BluetoothGattCharacteristic characteristic) {
+    public Observable<byte[]> readCharacteristic(@NonNull BluetoothGattCharacteristic characteristic) {
         return rxBleRadio.queue(new RxBleRadioOperationCharacteristicRead(
                 gattCallback,
                 bluetoothGatt,
@@ -265,23 +260,20 @@ public class RxBleConnectionImpl implements RxBleConnection {
     }
 
     @Override
-    public Observable<byte[]> writeCharacteristic(
-            @NonNull UUID characteristicUuid, @NonNull byte[] data) {
+    public Observable<byte[]> writeCharacteristic(@NonNull UUID characteristicUuid, @NonNull byte[] data) {
         return getCharacteristic(characteristicUuid)
                 .flatMap(characteristic -> writeCharacteristic(characteristic, data));
     }
 
     @Deprecated
     @Override
-    public Observable<BluetoothGattCharacteristic> writeCharacteristic(
-            @NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic) {
+    public Observable<BluetoothGattCharacteristic> writeCharacteristic(@NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic) {
         return writeCharacteristic(bluetoothGattCharacteristic, bluetoothGattCharacteristic.getValue())
                 .map(bytes -> bluetoothGattCharacteristic);
     }
 
     @Override
-    public Observable<byte[]> writeCharacteristic(
-            @NonNull BluetoothGattCharacteristic characteristic, @NonNull byte[] data) {
+    public Observable<byte[]> writeCharacteristic(@NonNull BluetoothGattCharacteristic characteristic, @NonNull byte[] data) {
         return rxBleRadio.queue(new RxBleRadioOperationCharacteristicWrite(
                 gattCallback,
                 bluetoothGatt,
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
@@ -63,7 +63,7 @@ public class RxBleConnectionImpl implements RxBleConnection {
     @Override
     public Observable<Integer> requestMtu(int mtu) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP)
-            return privateRequestMtu(mtu, 20, TimeUnit.SECONDS);
+            return privateRequestMtu(mtu, 10, TimeUnit.SECONDS);
         else
             return Observable.error(new BleException("requestMtu is not supported with API lower than 21"));
     }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
@@ -81,7 +81,6 @@ public class RxBleConnectionImpl implements RxBleConnection {
         return newObservable;
     }
 
-
     @Override
     public Observable<RxBleDeviceServices> discoverServices() {
         return privateDiscoverServices(20, TimeUnit.SECONDS);
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
@@ -4,12 +4,15 @@ import android.bluetooth.BluetoothGatt;
 import android.bluetooth.BluetoothGattCharacteristic;
 import android.bluetooth.BluetoothGattDescriptor;
 import android.bluetooth.BluetoothGattService;
+import android.os.Build;
 import android.support.annotation.NonNull;
+import android.support.annotation.RequiresApi;
 
 import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDeviceServices;
 import com.polidea.rxandroidble.exceptions.BleCannotSetCharacteristicNotificationException;
 import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetException;
+import com.polidea.rxandroidble.exceptions.BleException;
 import com.polidea.rxandroidble.internal.RxBleRadio;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicRead;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicWrite;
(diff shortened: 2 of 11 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -294,4 +294,32 @@ public String toString() {
...
+     * Timeouts after specified amount of time.
+     *
+     * @param timeout multiplier of TimeUnit after which the discovery will timeout in case of no return values
+     * @param timeUnit TimeUnit for the timeout
+     * @return Observable emitting result the MTU requested.
+     * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
+     *                                     the MTU for internal reasons.
+     * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
+     */
+    Observable<Integer> requestMtu(int mtu, long timeout, TimeUnit timeUnit);
```
dariuszseweryn: I am not really convinced if this functionality is a subject to a timeout - I have not seen any reports of a stalled negotiation of the MTU. It would be best not to have an overcomplicated interface if it is not needed.

pregno:
I've experienced some stalled negotiation when the device is not paired with the phone (or is not paired correctly). 
In that case, sometimes, you will never get a response.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -57,6 +58,36 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
...
+    public Observable<Integer> requestMtu(int mtu, long timeout, TimeUnit timeUnit) {
+        return privateRequestMtu(mtu, timeout, timeUnit);
+    }
+
+    private Observable<Integer> privateRequestMtu(int mtu, long timeout, TimeUnit timeUnit) {
+        synchronized (discoveredServicesCache) {
+            final Observable<Integer> newObservable;
+            // performing actual discovery
+                newObservable = rxBleRadio
+                        .queue(new RxBleRadioOperationMtuRequest(
```
uKL: Is it possible that the MTU will be changed by the device?

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
```diff
@@ -38,6 +45,29 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode,
+                                           @NonNull String[] permissions,
+                                           @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        switch (requestCode) {
+            case PERMISSIONS_REQUEST_LOCATION:
+                if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                    // permission was granted, yay! Do the
```
uKL: Our checkstyle cries because of empty statement. In this case it's not relevant but please satisfy it either by adding some action, removing some code or adding checkstyle ignore as this is an example only.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -294,4 +294,32 @@ public String toString() {
...
+     *
+     * @return Observable emitting result the MTU requested.
+     * @throws BleGattCannotStartException with {@link BleGattOperationType#ON_MTU_CHANGED} type, when it wasn't possible to set
+     *                                     the MTU for internal reasons.
+     * @throws BleGattException            in case of GATT operation error with {@link BleGattOperationType#ON_MTU_CHANGED} type.
+     */
+    Observable<Integer> requestMtu(int mtu);
+
+    /**
+     * Performs GATT request MTU
```
uKL: As far as I understand it requires API Level >=21 to use it. I think it would be convenient if the documentation would state that plus there may be an exception like `ApiNotSupportedException(int apiLevel)`. @dariuszseweryn what do you think?

dariuszseweryn:
@pregno could you find some information about the stalled negotiation somewhere on the stackoverflow or other place? I would like to read more about it.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
pregno: As far as I know, it's not possible.
The device I'm using automatically negotiate the MTU with iOS devices, but on Android this didn't happen

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
pregno: I don't usually write empty statements :)
I left it empty because I thought was not relevant for the sample.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
pregno: Yes, probably I've put `@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)` in the wrong place.
Right now is in the `protectedRun` method of the `RxBleRadioOperationMtuRequest` class

pregno:
Unfortunately not, but during my tests I've experienced when:
- device not paired
- device appears in the bonded list, but actually is not paired
- device goes out range when connecting

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
uKL: That's why putting checkstyle ignore would fit in this case :)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
uKL: I'm asking because if that's possible then with cache the method may return old values, which is no longer valid.

pregno: In that case we should put a listener (like the connection listener) on the callback `onMtuChanged`.
BTW I'll ask to my firmware developer to test that case

dariuszseweryn:
In this situation how did you tested it? Using what code?

pregno:
The tests has been done on my application, without using your library, so plain (and pain) Android BLE stack.The application is a companion app for our product, so we tested it intensely in every condition.

I know it is a borderline situation, maybe it depends on the firmware implementations.
If you are not convinced about that we could remove the timeout from the request and add it lately if  happens again

dariuszseweryn:
The situation with device going out of range will be automatically handled by the library as `RxBleGattCallback` is reporting an error if the device is disconnected.

About the case when the device is not bonded - I see no hint anywhere that negotiating MTU requires bonding. There are some issues regarding bonding in the Android OS but it's not directly connected with the MTU. If you know more about bonding / mtu - please say or direct me to a place where I can read more.

I think that having a programmable timeout is not the best idea (differently to discovering services as it will be longer if a peripheral has a lot of services / characteristics) because MTU is a two-packet procedure and shouldn't take long. The simpler the API is the better in my opinion.

pregno:
ok

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -96,7 +124,8 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
     }
 
     @Override
-    public Observable<BluetoothGattCharacteristic> getCharacteristic(@NonNull UUID characteristicUuid) {
```
dariuszseweryn: These formatting changes are not relevant to this commit.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
```diff
@@ -0,0 +1,52 @@
...
+        this.mtu = mtu;
+        this.rxBleGattCallback = rxBleGattCallback;
+        this.bluetoothGatt = bluetoothGatt;
+    }
+
+    @Override
+    protected void protectedRun() {
+        final Subscription subscription = rxBleGattCallback
+                .getOnMtuChanged()
+                .first()
```
dariuszseweryn: Having a `.timeout()` here is not a bad idea - it just doesn't have to be exposed at the interface. Plus a comment about why the timeout is left here could be useful to future maintainers.

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
dariuszseweryn: This part is also not directly bound to the MTU functionality and could be extracted to a different PR as it will be a benefit even without the MTU.

lowwor:
Can we have a snapshot version of this feature, since it's really useful and I really need this feature.
Or any other way I can achieve this function without compile the whole project as a dependency (It seems BluetoothGatt is not exposed).
Thanks in advance!

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
pregno: What timeout do you suggest? 20 seconds like for the discovery?

dariuszseweryn: I don't have any BLE device that supports non-standard MTU to test. 20 seconds looks like too much as service discovery is a long task where negotiation should be very quick. If you have a slow device with a less than optimal BLE stack (preferably some cheapish emerging market phone like Micromax Canvas) you could perform some tests how long does it usually take. Could you do the tests?

pregno: Unfortunately I can't test with slow devices. Our device is answering fast, even with the service discovery.

dariuszseweryn: Let's make it 10 seconds then. I think it could be even 5 but can't prove it without tests.

pregno: done

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -57,6 +61,33 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
     }
 
     @Override
+    public Observable<Integer> requestMtu(int mtu) {
```
dariuszseweryn: Would be good to have this one also annotated with `@RequiresApi`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -57,6 +61,33 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
...
+        newObservable = rxBleRadio
+                .queue(new RxBleRadioOperationMtuRequest(
+                        mtu,
+                        gattCallback,
+                        bluetoothGatt,
+                        timeout,
+                        timeUnit,
+                        Schedulers.computation()
+                ))
+                .cacheWithInitialCapacity(1);
```
dariuszseweryn: .cache is not used here basically

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
```diff
@@ -0,0 +1,70 @@
...
+        this.timeoutTimeUnit = timeoutTimeUnit;
+        this.timeoutScheduler = timeoutScheduler;
+    }
+
+    @Override
+    protected void protectedRun() {
+        final Subscription subscription = rxBleGattCallback
+                .getOnMtuChanged()
+                .first()
+                .timeout(timeout, timeoutTimeUnit, timeoutFallbackProcedure(), timeoutScheduler)
```
dariuszseweryn: timeoutFallbackProcedure is not needed here

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequestTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+        when:
+        changedMtuPublishSubject.onError(testException)
+
+        then:
+        testSubscriber.assertError(testException)
+
+        and:
+        (1.._) * mockSemaphore.release() // technically it's not an error to call it more than once
+    }
+
```
dariuszseweryn: Please add a test for timeout.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -294,4 +296,17 @@ public String toString() {
      * @return Observable emitting the read RSSI value
      */
     Observable<Integer> readRssi();
+
+    /**
+     * Performs GATT request MTU
+     *
+     * Timeouts after specified amount of time.
```
dariuszseweryn: Timeouts after 10 seconds.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
```diff
@@ -12,8 +12,8 @@
 import com.polidea.rxandroidble.exceptions.BleGattException;
 import com.polidea.rxandroidble.exceptions.BleGattOperationType;
 import com.polidea.rxandroidble.internal.RxBleLog;
-
```
dariuszseweryn: not relevant to the functionality

dariuszseweryn:
Thanks for re-adding the timeout as it is useful till we will deal with bonding related issues. I have added comments to the code.
Things to do before I can merge the changes:

1. Fix checkstyle issues.
2. Add a test for timeouting MTU negotiation operation.
3. Add a `RequiresApi` on `RxBleConnectionImpl.requestMtu(int mtu)` call.
4. Remove all changes that do not touch functionality of the mtu. (new lines in import).
5. Extract changes regarding getting the permissions to a different pull request.

Once the above things will be addressed we should be there. 👍

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequestTest.groovy
pregno: done

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
pregno: what is not relevant?

dariuszseweryn: removed line :) And added one line below

pregno: oh c'mon man, Android studio added

dariuszseweryn: From my experience any added / deleted line that is not directly connected to the change only obfuscates the code. Minimal changes are the best changes as it is easier to spot mistakes.

dariuszseweryn:
Three things left:

1. There are some obsolete changes with line breaks in `RxBleConnectionImpl`
2. Checkstyle is whining about unused imports in `RxBleRadioOperationMtuRequest` you can check it here: https://travis-ci.org/Polidea/RxAndroidBle/builds/181702296
3. There is now unused line in `ScanActivity` -> `public final static int PERMISSIONS_REQUEST_LOCATION = 16;`

I will mark them in the code to make it easier.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -107,7 +135,8 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
     }
 
     @Override
-    public Observable<Observable<byte[]>> setupNotification(@NonNull BluetoothGattCharacteristic characteristic) {
```
dariuszseweryn: this change with line break is obsolete

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -117,7 +146,8 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
     }
 
     @Override
-    public Observable<Observable<byte[]>> setupIndication(@NonNull BluetoothGattCharacteristic characteristic) {
```
dariuszseweryn: and this change with line break is obsolete

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -225,7 +255,8 @@ private void dismissTriggeredRead(
     }
 
     @Override
-    public Observable<byte[]> readCharacteristic(@NonNull BluetoothGattCharacteristic characteristic) {
```
dariuszseweryn: and this change with line break is obsolete

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -234,20 +265,23 @@ private void dismissTriggeredRead(
     }
 
     @Override
-    public Observable<byte[]> writeCharacteristic(@NonNull UUID characteristicUuid, @NonNull byte[] data) {
```
dariuszseweryn: and this change with line break is obsolete

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -234,20 +265,23 @@ private void dismissTriggeredRead(
...
+    public Observable<byte[]> writeCharacteristic(
+            @NonNull UUID characteristicUuid, @NonNull byte[] data) {
         return getCharacteristic(characteristicUuid)
                 .flatMap(characteristic -> writeCharacteristic(characteristic, data));
     }
 
     @Deprecated
     @Override
-    public Observable<BluetoothGattCharacteristic> writeCharacteristic(@NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic) {
```
dariuszseweryn: and this change with line break is obsolete

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -234,20 +265,23 @@ private void dismissTriggeredRead(
...
+    public Observable<BluetoothGattCharacteristic> writeCharacteristic(
+            @NonNull BluetoothGattCharacteristic bluetoothGattCharacteristic) {
         return writeCharacteristic(bluetoothGattCharacteristic, bluetoothGattCharacteristic.getValue())
                 .map(bytes -> bluetoothGattCharacteristic);
     }
 
     @Override
-    public Observable<byte[]> writeCharacteristic(@NonNull BluetoothGattCharacteristic characteristic, @NonNull byte[] data) {
```
dariuszseweryn: and this change with line break is obsolete

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
```diff
@@ -23,6 +28,8 @@
 
 public class ScanActivity extends AppCompatActivity {
 
+    public final static int PERMISSIONS_REQUEST_LOCATION = 16;
```
dariuszseweryn: line 31 and 32 are not needed as the permission request was extracted to a different PR

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
```diff
@@ -0,0 +1,65 @@
+package com.polidea.rxandroidble.internal.operations;
+
+import android.bluetooth.BluetoothGatt;
+import android.os.Build;
+import android.support.annotation.NonNull;
```
dariuszseweryn: This import is unused according to https://travis-ci.org/Polidea/RxAndroidBle/builds/181702296 checkstyle.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.polidea.rxandroidble.exceptions.BleGattCannotStartException;
+import com.polidea.rxandroidble.exceptions.BleGattOperationType;
+import com.polidea.rxandroidble.internal.RxBleRadioOperation;
+import com.polidea.rxandroidble.internal.connection.RxBleGattCallback;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import rx.Observable;
```
dariuszseweryn: This import is unused according to https://travis-ci.org/Polidea/RxAndroidBle/builds/181702296 checkstyle.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationMtuRequest.java
```diff
@@ -0,0 +1,65 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.RequiresApi;
+
+import com.polidea.rxandroidble.exceptions.BleGattCannotStartException;
+import com.polidea.rxandroidble.exceptions.BleGattOperationType;
+import com.polidea.rxandroidble.internal.RxBleRadioOperation;
+import com.polidea.rxandroidble.internal.connection.RxBleGattCallback;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
dariuszseweryn: This import is unused according to https://travis-ci.org/Polidea/RxAndroidBle/builds/181702296 checkstyle.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -4,17 +4,21 @@
...
 import android.bluetooth.BluetoothGattService;
+import android.os.Build;
 import android.support.annotation.NonNull;
+import android.support.annotation.RequiresApi;
 
 import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDeviceServices;
 import com.polidea.rxandroidble.exceptions.BleCannotSetCharacteristicNotificationException;
 import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetException;
+import com.polidea.rxandroidble.exceptions.BleException;
```
dariuszseweryn: This import is unused according to https://travis-ci.org/Polidea/RxAndroidBle/builds/181702296 checkstyle.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -57,6 +61,30 @@ public RxBleConnectionImpl(RxBleRadio rxBleRadio, RxBleGattCallback gattCallback
...
     @Override
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    public Observable<Integer> requestMtu(int mtu) {
+        return privateRequestMtu(mtu, 10, TimeUnit.SECONDS);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private Observable<Integer> privateRequestMtu(int mtu, long timeout, TimeUnit timeUnit) {
+        final Observable<Integer> newObservable;
+        // performing actual discovery
```
dariuszseweryn: this comment is not needed (probably a copy-paste error)

dariuszseweryn:
@pregno I hope that you are not taking my requests personally. Our current setup is currently not building artifacts if the `checkstyle` is complaining - thus the need of removing unused imports. And it doesn't look good if there are changes that are not relevant to the topic of a pull request - so removing unnecessary formatting and leftovers from parts that were already removed will boil down the code to easily understandable bare minimum. 
I am really thankful to you for committing your work towards what I hope will be the best android ble library. :+1:

pregno:
Don't worry :)
I will close the various fixes soon

sample/src/main/java/com/polidea/rxandroidble/sample/example1_scanning/ScanActivity.java
```diff
@@ -23,6 +23,7 @@
 
 public class ScanActivity extends AppCompatActivity {
 
+
```
dariuszseweryn: this is not needed

pregno:
ok, done

dariuszseweryn:
Thank you for your effort @pregno - your changes are merged and will be available in the 1.2.0-SNAPSHOT release shortly. :) 💯 
Best Regards

pregno:
thank you too!

best

lowwor:
Great job! Thank you all for this nice feature.

PR #105, opened 2016-12-06
Avoid forcing the value for autoConnect

JlUgia:
If the first condition is met, autoConnect has the right to be both `true` or `false`.


uKL:
I think it's still invalid. I think it should be Build.VERSION.SDK_INT < Build.VERSION_CODES.N && autoConnect

Dariusz, what do you think? The compatible should be applied to pre-made when autoConnect == true.

dariuszseweryn:
I think that @JlUgia is right.
The condition seems right. If the `Build.VERSION.SDK_INT` is newer than or equal to Nougat then we're not using reflection. If the `autoConnect` is false we are also not using reflection.

uKL:
That's right. I didn't see whole file on my mobile. I thought that reflection is in the `connectGattCompat`. I'd merge this one.

Shouldn't we have some tests for this logic?

dariuszseweryn:
We could think about adding some tests but this one is ready for merge right away. Then we need to make a release of `1.1.1`

uKL:
Dear Contributor,

PR #112, opened 2017-01-03
Added requestConnectionPriority methods.

svangsgaard:
I've tried to implement requestConnectionPriority support.
Thank you for an awesome library.

### DIFF
--- a/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
+++ b/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
@@ -43,6 +43,11 @@ class RxBleConnectionMock implements RxBleConnection {
     }
 
     @Override
+    public void requestConnectionPriority(int priority) {
+
+    }
+
+    @Override
     public Observable<RxBleDeviceServices> discoverServices() {
         return Observable.just(rxBleDeviceServices);
     }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -17,6 +17,7 @@ import com.polidea.rxandroidble.exceptions.BleGattOperationType;
 import java.util.UUID;
 
 import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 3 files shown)


dariuszseweryn:
Hello @svangsgaard 

Thank you for your pull request - this is a great idea! Unfortunately I cannot merge it yet, there are few things that need to be clarified:

1. Setting the connection priority may fail and your code does not inform the user about the result
2. Have you tested wether requesting connection priority may be called at any time or does it have to be called when no bluetooth operation is executed at that time? If it is the case we would need to make the request asynchronous and create a RxBleRadioOperationRequestConnectionPriority and related tests
3. It may be also worth to discuss if a High Priority or Low Priority connection would have a lifecycle similar to normal connection and / or notifications. I.E. when user will request high priority connection it will be returned in `Observable<RxBleConnectionHighPriority>` which would have almost the same interface as `RxBleConnection` but would not be able to change the priority. When the user would unsubscribe from the `Observable<RxBleConnectionHighPriority>` the priority would be brought back to balanced state. What is your use-case?
4. Your commit introduces formatting changes in JavaDoc which are not related to the topic and should be reverted.

After all I am looking forward to have this feature after we will discuss and implement it in a proper way.

Best Regards

svangsgaard:
Hi @dariuszseweryn 

You're right. I need to change those things in the code.
I have thought about how to implement it, maybe a better solution is to provide an extra parameter to setupNotification aka setupNotification(MY_UUID, CONNECTION_PRIORITY_HIGH), then the priority is automatically reverted when the notification is released.

That will solve my current problem. What do you think?

Regards.

dariuszseweryn:
I don't think that binding connection priority to setting up notifications is a good idea. I can imagine at least a situation where user would like to transmit data to the peripheral at increased rate.

Could you describe your use case in greater detail? Does your peripheral transmit a lot of data via notifications to the phone?

I am on the fence if introducing state to the connection via setting the connection priority is a good idea. Maybe it would be better to emit `RxBleConnectionHighPriority` but what if at the same time the user would also request low priority? Then an error should be emitted.

Either way - if you could confirm the point number 2 from my previous post - it would be clear if we need a separate operation.

dariuszseweryn:
@svangsgaard Any hints you can provide?

mikolak:
Now implemented. https://github.com/Polidea/RxAndroidBle/issues/111

PR #139, opened 2017-02-21
Added a way to execute custom RxBleRadioOperation

maoueh:
The RxBleConnection interface has now a new method named queue that enable
library consumer to queue their own RxBleOperation. The queue method
receives in parameter a builder function taking three arguments. In order,
the RxBleRadio scheduler (not required to be used), the BluetoothGatt object
to interact with Android BLE API and RxBleGattCallback to receive Android
BLE API callbacks.

The builder parameter should return a concrete implementation of
RxBleOperation that will be executed as part of normal RxBleRadio operation.
This custom RxBleOperation must respect the RxBleOperation contract by
releasing the semaphore, otherwise, queue will be stuck forever.

This feature shall be used only by people that studied the RxAndroidBLE
source code and understand the implication of running their own operation.

### DIFF
--- a/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
+++ b/mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
@@ -7,6 +7,7 @@ import android.support.annotation.NonNull;
 import com.polidea.rxandroidble.NotificationSetupMode;
 import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDeviceServices;
+import com.polidea.rxandroidble.RxBleRadioOperationCustom;
 import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetException;
 import com.polidea.rxandroidble.internal.connection.ImmediateSerializedBatchAckStrategy;
 import com.polidea.rxandroidble.internal.util.ObservableUtil;
@@ -164,7 +165,7 @@ public class RxBleConnectionMock implements RxBleConnection {
 
     @Override
     public Observable<Observable<byte[]>> setupIndication(@NonNull UUID characteristicUuid) {
-       return setupIndication(characteristicUuid, NotificationSetupMode.DEFAULT);
+        return setupIndication(characteristicUuid, NotificationSetupMode.DEFAULT);
     }
 
     @Override
@@ -467,5 +468,10 @@ public class RxBleConnectionMock implements RxBleConnection {
             return just(true);
         }
 
+    }
+
+    @Override
+    public <T> Observable<T> queue(RxBleRadioOperationCustom<T> operation) {
+        throw new UnsupportedOperationException("Mock does not support queuing custom operation.");
     }
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
@@ -1,5 +1,6 @@
 package com.polidea.rxandroidble;
 
+import android.bluetooth.BluetoothGatt;
 import android.bluetooth.BluetoothGattCharacteristic;
 import android.bluetooth.BluetoothGattDescriptor;
 import android.content.Context;
@@ -13,11 +14,13 @@ import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetE
 import com.polidea.rxandroidble.exceptions.BleGattCannotStartException;
 import com.polidea.rxandroidble.exceptions.BleGattException;
 import com.polidea.rxandroidble.exceptions.BleGattOperationType;
+import com.polidea.rxandroidble.internal.connection.RxBleGattCallback;
 
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 import rx.Observable;
+import rx.Scheduler;
 
 /**
  * The BLE connection handle, supporting GATT operations. Operations are enqueued and the library makes sure that they are not
@@ -417,4 +420,33 @@ public interface RxBleConnection {
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     Observable<Integer> requestMtu(int mtu);
+
+    /**
+     * <b>This method requires deep knowledge of RxAndroidBLE internals. Use it only as a last resort if you know
+     * what your are doing.</b>
+     * <p>
+     * Queue an operation for future execution. The method accepts a {@link RxBleRadioOperationCustom} concrete implementation
+     * and will queue it inside connection operation queue. When ready to execute, the {@link Observable<T>} returned
+     * by the {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)} will be
+     * subscribed to.
+     * <p>
+     * Every event emitted by the {@link Observable<T>} returned by
+     * {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)} will be forwarded
+     * to the {@link Observable<T>} returned by this method.
+     * <p>
+     * You <b>must</b> ensure the custom operation's {@link Observable<T>} do terminate either via {@code onCompleted}
+     * or {@code onError(Throwable)}. Otherwise, the internal queue orchestrator will wait forever for
+     * your {@link Observable<T>} to complete. Normal queue processing will be resumed after the {@link Observable<T>}
+     * returned by {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)}
+     * completes.
+     * <p>
+     * The operation will be added to the queue using a {@link com.polidea.rxandroidble.internal.RxBleRadioOperation.Priority#NORMAL}
+     * priority.
+     *
+     * @param operation The custom radio operation to queue.
+     * @param <T>       The type returned by the {@link RxBleRadioOperationCustom} instance.
+     * @return Observable emitting the value after execution or an error in case of failure.
+     */
+    <T> Observable<T> queue(RxBleRadioOperationCustom<T> operation);
+
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
@@ -1,0 +1,50 @@
+package com.polidea.rxandroidble;
+
+import android.bluetooth.BluetoothGatt;
+import android.support.annotation.NonNull;
+
+import com.polidea.rxandroidble.internal.RxBleRadio;
+import com.polidea.rxandroidble.internal.connection.RxBleGattCallback;
+
+import rx.Observable;
+import rx.Observer;
+import rx.Scheduler;
+import rx.android.schedulers.AndroidSchedulers;
+
+/**
+ * Represents a custom operation that will be enqueued for future execution within the client instance.
+ */
+public interface RxBleRadioOperationCustom<T> {
+
+    /**
+     * Return an observable that implement a custom radio operation using low-level Android BLE API.
+     * <p>
+     * The {@link Observable<T>} returned by this method will be subscribed to by the {@link RxBleRadio}
+     * when it determines that the custom operation should be the next to be run.
+     * <p>
+     * The method receives everything needed to access the low-level Android BLE API objects mainly the
+     * {@link BluetoothGatt} to interact with Android BLE GATT operations and {@link RxBleGattCallback}
+     * to be notified when GATT operations completes.
+     * <p>
+     * Every event emitted by the returned {@link Observable<T>} will be forwarded to the observable
+     * returned by {@link RxBleConnection#queue(RxBleRadioOperationCustom)}
+     * <p>
+     * As the implementer, your contract is to return an {@link Observable<T>} that completes at some
+     * point in time. When the returned observable terminates, either via the {@link Observer#onCompleted()} or
+     * {@link Observer#onError(Throwable)} callback, the {@link RxBleRadio} queue's lock is released so that
+     * queue operations can continue.
+     * <p>
+     * You <b>must</b> ensure the returned {@link Observable<T>} do terminate either via {@code onCompleted}
+     * or {@code onError(Throwable)}. Otherwise, the internal queue orchestrator will wait forever for
+     * your {@link Observable<T>} to complete and the it will not continue to process queued operations.
+     *
+     * @param bluetoothGatt     The Android API GATT instance
+     * @param rxBleGattCallback The internal Rx ready bluetooth gatt callback to be notified of GATT operations
+     * @param scheduler         The RxBleRadio scheduler used to asObservable operation (currently {@link AndroidSchedulers#mainThread()}
+     * @throws Throwable Any exception that your custom operation might throw
+     */
+    @NonNull
+    Observable<T> asObservable(BluetoothGatt bluetoothGatt,
+                               RxBleGattCallback rxBleGattCallback,
+                               Scheduler scheduler) throws Throwable;
+}
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadio.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadio.java
@@ -1,8 +1,11 @@
 package com.polidea.rxandroidble.internal;
 
 import rx.Observable;
+import rx.Scheduler;
 
 public interface RxBleRadio {
+
+    Scheduler scheduler();
 
     <T> Observable<T> queue(RxBleRadioOperation<T> rxBleRadioOperation);
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadioOperation.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadioOperation.java
@@ -50,7 +50,7 @@ public abstract class RxBleRadioOperation<T> implements Runnable, Comparable<RxB
     }
 
     /**
-     * This method will be overriden in concrete operation implementations and
+     * This method will be overridden in concrete operation implementations and
      * will contain specific operation logic.
      */
     protected abstract void protectedRun() throws Throwable;
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
@@ -5,15 +5,20 @@ import android.bluetooth.BluetoothGattCharacteristic;
 import android.bluetooth.BluetoothGattDescriptor;
 import android.bluetooth.BluetoothGattService;
 import android.os.Build;
+import android.os.DeadObjectException;
 import android.support.annotation.NonNull;
 import android.support.annotation.RequiresApi;
 
 import com.polidea.rxandroidble.NotificationSetupMode;
 import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.RxBleDeviceServices;
+import com.polidea.rxandroidble.RxBleRadioOperationCustom;
 import com.polidea.rxandroidble.exceptions.BleCannotSetCharacteristicNotificationException;
 import com.polidea.rxandroidble.exceptions.BleConflictingNotificationAlreadySetException;
+import com.polidea.rxandroidble.exceptions.BleDisconnectedException;
+import com.polidea.rxandroidble.exceptions.BleException;
 import com.polidea.rxandroidble.internal.RxBleRadio;
+import com.polidea.rxandroidble.internal.RxBleRadioOperation;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicRead;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationCharacteristicWrite;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationDescriptorRead;
@@ -22,8 +27,8 @@ import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationMtuReques
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationReadRssi;
 import com.polidea.rxandroidble.internal.operations.RxBleRadioOperationServicesDiscover;
 import com.polidea.rxandroidble.internal.util.ByteAssociation;
-import com.polidea.rxandroidble.internal.util.CharacteristicNotificationId;
 import com.polidea.rxandroidble.internal.util.CharacteristicChangedEvent;
+import com.polidea.rxandroidble.internal.util.CharacteristicNotificationId;
 import com.polidea.rxandroidble.internal.util.ObservableUtil;
 
 import java.util.HashMap;
@@ -527,5 +532,31 @@ public class RxBleConnectionImpl implements RxBleConnection {
     @Override
     public Observable<Integer> readRssi() {
         return rxBleRadio.queue(new RxBleRadioOperationReadRssi(gattCallback, bluetoothGatt, timeoutScheduler));
+    }
+
+    @Override
+    public <T> Observable<T> queue(final RxBleRadioOperationCustom<T> operation) {
+        return rxBleRadio.queue(new RxBleRadioOperation<T>() {
+            @Override
+            @SuppressWarnings("ConstantConditions")
+            protected void protectedRun() throws Throwable {
+                Observable<T> operationObservable = operation.asObservable(bluetoothGatt, gattCallback, rxBleRadio.scheduler());
+                if (operationObservable == null) {
+                    throw new IllegalArgumentException("The custom operation asObservable method must return a non-null observable");
+                }
+
+                operationObservable.doOnCompleted(new Action0() {
+                    @Override
+                    public void call() {
+                        releaseRadio();
+                    }
+                }).subscribe(getSubscriber());
+            }
+
+            @Override
+            protected BleException provideException(DeadObjectException deadObjectException) {
+                return new BleDisconnectedException(deadObjectException, bluetoothGatt.getDevice().getAddress());
+            }
+        });
     }
 }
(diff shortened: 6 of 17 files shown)


mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
```diff
@@ -468,4 +473,9 @@ public Boolean call(byte[] ignored) {
         }
 
     }
+
+    @Override
+    public <T> Observable<T> queue(Func3<Scheduler, BluetoothGatt, RxBleGattCallback, RxBleRadioOperation<T>> builder) {
+        return Observable.never();
```
maoueh: Was not sure what to do for this, should mock support custom operation?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadio.java
```diff
@@ -1,8 +1,11 @@
 package com.polidea.rxandroidble.internal;
 
 import rx.Observable;
+import rx.Scheduler;
 
 public interface RxBleRadio {
 
+    Scheduler scheduler();
```
maoueh: This was done mainly so that custom operation can access the scheduler (via the builder). The `LongWriteOperationBuilderImpl` now also use `RxBleRadio#scheduler` instead of passing around `AndroidSchedulers.mainThread()`.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -417,4 +423,23 @@ public String toString() {
...
+     * <li>{@link Scheduler} The RxBleRadio scheduler used to execute operation ({@link AndroidSchedulers#mainThread()}</li>
+     * <li>{@link BluetoothGatt} The Android API GATT instance</li>
+     * <li>{@link RxBleGattCallback} The internal Rx ready bluetooth gatt callback to be notified of GATT operations</li>
+     * </ol>
+     *
+     * @param builder The builder used to create the {@link RxBleRadioOperation} instance
+     * @param <T>     The type returned by the {@link RxBleRadioOperation} instance
+     * @return Observable emitting the value after execution or an error in case of failure.
+     */
+    <T> Observable<T> queue(Func3<Scheduler, BluetoothGatt, RxBleGattCallback, RxBleRadioOperation<T>> builder);
```
dariuszseweryn: I have few thoughts:

1. If we would additionally pass an `Action0 releaseRadioAction` - which would be a proxy to `releaseRadio()` method - it would be easier to explain that calling the `releaseRadioAction` is needed to resume a normal radio operations
2. If we would pass a `Observer<T> observerProxy` it would be easy to describe that everything that goes into the `observerProxy` will be available in the returned `Observable<T>`
3. Maybe the builder pattern here is not needed indirection? Maybe it will be better to create an abstract class like `RxBleRadioOperationCustom` and only allow to run subclasses of it?

mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
maoueh: I look at it a little but it seems hard to support it mainly because there is no `BluetoothGatt` nor `RxBleGattCallback` instances available that I could pass to custom operation.

This means we would need to some mock for them (which might be hard for `BluetoothGatt`) or pass `null` which would not exercise the custom operation.

I think people should instead test them the way you guys do in the RxAndroidBLE code base.

As such, I would vote for throwing an exception, probably `UnsupportedOperationException`.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
maoueh: I'm all in for 1 and 2, I thought about passing them, but wasn't sure, will change.

For 3, do you see `RxBleRadioOperationCustom` as `RxBleRadioOperationCustom extends RxBleRadioOperation` or as a standalone class (would probably be an interface to me).

I see it as this (I would like to come up with a better name than `RxBleRadioOperationCustom` however):

```
    interface RxBleRadioOperationCustom<T> {
        void execute(BluetoothGatt bluetoothGatt,
                            RxBleGattCallback rxBleGattCallback,
                            Scheduler scheduler,
                            Observer<T> observerProxy,
                            Action0 releaseRadioAction) {
        } throws Exception
    }
```

Then the `queue` would adapt this interface to an `RxBleRadioOperation`:

```
    @Override
    public <T> Observable<T> queue(final RxBleRadioOperationCustom<T> operation) {
        return rxBleRadio.queue(new RxBleRadioOperation<T>() {
            @Override
            protected void protectedRun() throws Throwable {
                operation.execute(bluetoothGatt, gattCallback, rxBleRadio.scheduler(), asObserver(), new Action0() {
                    @Override
                    public void call() {
                        releaseRadio();
                    }
                });
            }

            @Override
            protected BleException provideException(DeadObjectException deadObjectException) {
                return new BleDisconnectedException(deadObjectException, bluetoothGatt.getDevice().getAddress());
            }
        });
    }
```

The gain I see here is that `RxBleRadioOperation` is not exposed to consumer of the library meaning you are still free to refactor it as you see fit. You only need to keep backward-compatibility on the `RxBleRadioOperationCustom` interface now.

What do you think?

mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
dariuszseweryn: Mocking both `BluetoothGatt` and `RxBleGattCallback` should be totally doable. But adding proper support for it would take some time - I am not even sure how many users the mock library has to judge if it is worth the effort or not.
`UnsupportedOperationException` should be fine for now.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
dariuszseweryn: Agree - better to use an interface instead.
Best would be to have an interface with a single entry point as you have proposed because it could be lambdafied although then the custom operation would not be able to specify the `Priority`. A default implementation of a method from Java 8 would be a solution here but we need something different.

maoueh: As fro priority, what about:

```
    <T> Observable<T> queue(RxBleRadioOperationCustom<T> operation);

    <T> Observable<T> queue(RxBleRadioOperationCustom<T> operation, RxBleRadioOperation.Priority priority);
```

Where first version you defaults to `NORMAL`? Only drawback I see is that `RxBleRadioOperation.Priority` is kind of internal right now so, should I expose it publicly or keep it in internal package?

Thoughts?

mockrxandroidble/src/main/java/com/polidea/rxandroidble/mockrxandroidble/RxBleConnectionMock.java
maoueh: Good, will do that.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
maoueh: @dariuszseweryn I think the `Action0 releaseRadioAction` is not needed anymore. Indeed, since I know control the `RxBleRadioOperation` class now, we could release the radio once the operation has finished executed in a ` try/finally` block:

```
            @Override
            protected void protectedRun() throws Throwable {
                try {
                    operation.execute(bluetoothGatt, gattCallback, rxBleRadio.scheduler(), asObserver());
                } finally {
                    releaseRadio();
                }
            }
```

This seems even more fool-proof, what do you think?

dariuszseweryn: On the second thought it would be yet another method in the interface which I have imagined at the beginning to be as simple as possible.
We can stick with only normal priority for now. We can always add the additional function later when someone will signal such a need. That would only need a minor version bump and not too much other work anyway.

dariuszseweryn: As for the `try / finally` - Interesting idea.
Unfortunately I do not think it is a good idea since then it is not certain if other queued operations would not be executed while the custom operation is running. This could lead to clashes if the custom operation executes multiple calls on `BluetoothGatt` - like in your update firmware situation.
Great that you think about other possible solutions!

maoueh: > is not certain if other queued operations would not be executed while the custom operation is running

How come? If the `releaseRadio` is not called, no other queue operation can run. Unless I'm missing something, it's impossible to run other BLE operation while the custom one is running. It's the point of the queue, no?

dariuszseweryn: True. 
But the queue is continuing to be processed when the `releaseRadio()` is called. In the above snippet - to be sure that `releaseRadio()` is not called before the custom operation is finished - the `execute()` would need to be synchronous. And most of the operations on Android BLE are asynchronous - this is not really convenient. Moreover blocking in the `.execute()` would be on the main thread - a no-no.

maoueh: Yep, I see what you mean now, I didn't think about this fact. Seems it's needed ... unless `RxBleRadioOperationCustom#execute` returns an `Observable<T>` where I could subscribe and know when it's finish. That would also remove the need for the `Observer<T>` variable.

Not sure it's worth it however,  what do you think?

maoueh: I updated the PR to show the `Observable` idea and I added an example of a custom operation in the sample project.

Now that I think about it, not blocking the main thread might be a problem for people implementing custom operation. For example, let's assume my flashing firmware use case.

If I do `flashFirmareObservable().subcribeOn(scheduler);`, this will block the main thread for the whole flashing progress. To avoid this, I would need to at various interval to let a chance for the main thread to run something like:

```
  // The firmwareChunkWritters would be Observable<Observable<T>>
  return firmwareChunkWritters().concatMap(writter -> writter.subscribeOn(scheduler));
```

We cannot do anything I guess except warning user in the `RxBleRadioOperationCustom` documentation and in the sample.

Anyway, tell me what you think about the `Observable` approach. If you see a problem or you don't like it, will revert back to `observerProxy/releasedAction` way.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -417,4 +422,23 @@ public String toString() {
...
+     * <p>
+     * Queue an operation for future execution. The method accepts a {@link RxBleRadioOperationCustom} concrete implementation
+     * and will queue it inside connection operation queue. When ready to execute, the
+     * {@link RxBleRadioOperationCustom#execute(BluetoothGatt, RxBleGattCallback, Scheduler, Observer)} will be
+     * invoked.
+     * <p>
+     * The operation will be added to the queue using a {@link com.polidea.rxandroidble.internal.RxBleRadioOperation.Priority#NORMAL}
+     * priority.
+     *
+     * @param operation The builder used to create the {@link RxBleRadioOperation} instance
```
dariuszseweryn: the javadoc is outdated

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
```diff
@@ -0,0 +1,37 @@
...
+public interface RxBleRadioOperationCustom<T> {
+
+    /**
+     * Execute the custom operation. This method is called once the {@link com.polidea.rxandroidble.internal.RxBleRadio}
+     * is determining that this operation should be the next to be run.
+     * <p>
+     * The method receives everything needed to access the low-level Android BLE API objects.
+     * <p></p>
+     * As the implementer, you access {@link BluetoothGatt} and {@link RxBleGattCallback} to implement
+     * a custom operation and you use the {@link Observer<T>} {@code observerProxy} object to forward
```
dariuszseweryn: javadoc is outdated

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
```diff
@@ -0,0 +1,37 @@
...
+     * events to {@link rx.Observable<T>} return by the {@link RxBleConnection#queue(RxBleRadioOperationCustom)}
+     * operation. You <b>must</b> call {@link Observer#onCompleted() or {@link Observer#onError(Throwable)} otherwise
+     * the {@link rx.Observable<T>} will never really terminate.
```
dariuszseweryn: will never really terminate and the operations queue will not continue to process next events.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleRadioOperation.java
```diff
@@ -37,6 +38,15 @@
         return replaySubject;
     }
 
+    /**
+     * A function that returns this operation as an {@link Observer<T>} which will be forwarded to the
+     * {@link com.polidea.rxandroidble.RxBleRadioOperationCustom<T>} so that it is able to forward event
+     * to people subscribing to {@link #asObservable()}.
+     */
+    protected Observer<T> asObserver() {
```
dariuszseweryn: this is no longer needed - I cannot see any usage of it.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
```diff
@@ -655,6 +657,29 @@ class RxBleConnectionTest extends GradleRoboSpecification {
...
+        new RxBleRadioOperationCustom<Boolean>() {
+            @Override
+            Observable<Boolean> execute(BluetoothGatt bluetoothGatt,
+                               RxBleGattCallback rxBleGattCallback,
+                               Scheduler scheduler) throws Throwable {
+                Observable.just(true, false)
+            }
+        }
+    }
+
```
dariuszseweryn: We could have more tests:
"should release the radio if custom operation will throw out of RxBleCustomOperation.execute()"
"should release the radio if observable returned from RxBleCustomOperation.execute() will emit error"
"should release the radio when observable returned from RxBleCustomOperation.execute() will complete"
"should pass error if observable returned from RxBleCustomOperation.execute() will emit error"
"should pass items emitted by observable returned from RxBleCustomOperation.execute()"
Maybe there are even more scenarios that should be covered.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.polidea.rxandroidble.internal.connection.RxBleGattCallback;
+
+import rx.Observable;
+import rx.Observer;
+import rx.Scheduler;
+import rx.android.schedulers.AndroidSchedulers;
+
+/**
+ * Represents a custom operation that will be enqueue for future execution within the client instance.
```
uKL: enqueued

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -528,4 +533,32 @@ public BluetoothGattCharacteristic call(byte[] bytes) {
...
         return rxBleRadio.queue(new RxBleRadioOperationReadRssi(gattCallback, bluetoothGatt, timeoutScheduler));
     }
+
+    @Override
+    public <T> Observable<T> queue(final RxBleRadioOperationCustom<T> operation) {
+        return rxBleRadio.queue(new RxBleRadioOperation<T>() {
+            @Override
+            protected void protectedRun() throws Throwable {
+                try {
+                    operation.execute(bluetoothGatt, gattCallback, rxBleRadio.scheduler())
```
uKL: The operation is a public API, I think we should validate whether it behaves as we expect and check if returned value in not null.

Result of the assertion should clearly state the problem.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -528,4 +533,32 @@ public BluetoothGattCharacteristic call(byte[] bytes) {
     public Observable<Integer> readRssi() {
         return rxBleRadio.queue(new RxBleRadioOperationReadRssi(gattCallback, bluetoothGatt, timeoutScheduler));
     }
+
+    @Override
+    public <T> Observable<T> queue(final RxBleRadioOperationCustom<T> operation) {
+        return rxBleRadio.queue(new RxBleRadioOperation<T>() {
```
uKL: I think the default operation timeout should also apply to custom operations.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -528,4 +533,32 @@ public BluetoothGattCharacteristic call(byte[] bytes) {
...
+                            .doOnTerminate(new Action0() {
+                                @Override
+                                public void call() {
+                                    releaseRadio();
+                                }
+                            })
+                            .subscribe(getSubscriber());
+                } catch (Throwable throwable) {
+                    releaseRadio();
+
```
uKL: Remove blank line please.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
maoueh: Sorry, should have warned you that I did not update everything because I didn't want to invest time updating everything if `Observable<T>` was "rejected". I will fix all issues.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
maoueh: This would be wrong IMO, I'm flashing a firmware, I don't want to have a maximum amount of time to perform the operation as the default timeout could not fit what I need to do. 

We would need a rather larger timeout which would defeat the purpose.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
maoueh: Will do.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
maoueh: Ok

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
dariuszseweryn: I agree that the timeout is highly dependent on a use-case. Same problem could with the long write op which checks for timeouts internally

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -417,4 +420,27 @@ public String toString() {
...
+     * and will queue it inside connection operation queue. When ready to execute, the {@link Observable<T>} returned
+     * by the {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)} will be
+     * subscribed to.
+     * <p>
+     * Every event emitted by the {@link Observable<T>} returned by
+     * {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)} will be forwarded
+     * to the {@link Observable<T>} returned by this method.
+     * <p>
+     * The operation will be added to the queue using a {@link com.polidea.rxandroidble.internal.RxBleRadioOperation.Priority#NORMAL}
+     * priority.
```
dariuszseweryn: I would add `Normal queue processing will be resumed after the Observable returned by {@link RxBleRadioOperationCustom#asObservable(BluetoothGatt, RxBleGattCallback, Scheduler)} will complete`

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
```diff
@@ -0,0 +1,48 @@
...
+     * You <b>must</b> ensure the returned {@link Observable<T>} do terminate either via {@code onCompleted}
+     * or {@code onError(Throwable)}. Otherwise, the internal queue orchestrator will wait forever for
+     * your {@link Observable<T>} to complete and the it will not continue to process queued operations.
+     *
+     * @param bluetoothGatt     The Android API GATT instance
+     * @param rxBleGattCallback The internal Rx ready bluetooth gatt callback to be notified of GATT operations
+     * @param scheduler         The RxBleRadio scheduler used to asObservable operation (currently {@link AndroidSchedulers#mainThread()}
+     * @throws Throwable Any exception that your custom operation might throw
+     */
+    Observable<T> asObservable(BluetoothGatt bluetoothGatt,
```
dariuszseweryn: Maybe add a `@NonNull` annotation?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
uKL: Is there anything in terms of locking the radio permanently we should add to the JavaDoc then? Possibly if this observable wouldn't emit, we will get stuck. This is not so obvious for someone reading the documentation.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
```diff
@@ -655,6 +657,83 @@ class RxBleConnectionTest extends GradleRoboSpecification {
...
+                throw new RuntimeException()
+            }
+        }
+
+        when:
+        objectUnderTest.queue(operationBuilder).subscribe(testSubscriber)
+
+        then:
+        testSubscriber.assertError(RuntimeException.class)
+        flatRadio.semaphore.isReleased() >> true
```
uKL: Why did you use `flatRadio.semaphore.isReleased() >> true` here? flatRadio is not a mock. I think you can skip ` >> true` this part.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
```diff
@@ -655,6 +657,83 @@ class RxBleConnectionTest extends GradleRoboSpecification {
...
+
+    def "should pass error if observable returned from RxBleRadioOperationCustom.asObservable() will emit error"() {
+        given:
+        def operationBuilder = createCustomOperationBuilder(Observable.error(new RuntimeException()))
+
+        when:
+        objectUnderTest.queue(operationBuilder).subscribe(testSubscriber)
+
+        then:
+        testSubscriber.assertError(RuntimeException.class)
```
uKL: Shouldn't the semaphore be released as well?

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleRadioOperationCustom.java
maoueh: Yep, sure.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
maoueh: I was not very familiar with Spock and thought it was a way to do assertions, will remove.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
maoueh: Yeah, but it was already covered by another test case so I decided not to repeat. Will merge the two tests together I think.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
maoueh: Added + a sentence put emphasis on the fact the returned observable must complete.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
maoueh: The javadoc of both `RxBleRadioOperationCustom#asObservable` and `RxBleConnection#queue` (this is not pushed yet, I've just added it). I think it will be good now.

maoueh:
Updated with latest comments. I'm going to test the PR in "real situation" really soon.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/RxBleConnectionTest.groovy
maoueh: The previous thing was even more problematic as there was a double release bug and with the `flatRadio.semaphore.isReleased() >> true`, well it was not triggering correctly.

Thanks!

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
uKL: Yeah, I think so as well :)

maoueh:
@dariuszseweryn @uKL Tested this branch (and long write along the way) in my app using an LG G5 (API 24) and a Samsung S4 Mini (API 19), both worked correctly and did not find any issue related to long write and custom operation.

Still gathering numbers to show to you guys but it's taking more time as I'm being side-tracked by other things. So, I wanted to report back quickly at least that everything seems to be working great.

I would like to provide a "flash firmware" custom operation snippet somehow, where do you propose me to put this?

Ready to merge from my perspective, thanks for all the help and great suggestions.

uKL:
You can add an example to the sample project. It's up to you to open another PR or add it to this one. I prefer the first one as we have another change waiting to be merged after yours. :)

dariuszseweryn:
@maoueh In this situation - merged. Feel free to add another pull request with a sample of firmware update. 👍 
Thank you for your great work!

PR #181, opened 2017-04-26
Update build for gradle plugin 2.5

IgorGanapolsky:
Upgrade to gradle plugin 2.5 for the new build system.
Use Java 8 in Android Studio properly.
Get rid of unnecessary dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,10 +7,9 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.5.0-alpha-preview-02'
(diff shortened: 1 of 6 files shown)


uKL:
Is there any reason for this change? Does it fix anything?

IgorGanapolsky:
It adheres to:
**https://developer.android.com/guide/platform/j8-jack.html**
and
**http://tools.android.com/tech-docs/new-build-system/2-5-alpha-gradle-plugin**

uKL:
It's not likely that we will go into the alpha version unless it fixes something important. In this case:
- Jack toolchain is deprecated
- We want to have API support 18+

Let's wait for more stable version.

PR #184, opened 2017-04-27
Makes UUIDUtil constructor public to solve Kotlin problem

marciogranzotto:
Android Studio, while using Kotlin, was complaining that couldn't initialize the `UUIDUtil` class because the constructor wasn't `public`.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/UUIDUtil.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/UUIDUtil.java
@@ -20,7 +20,7 @@ import javax.inject.Inject;
 public class UUIDUtil {
 
     @Inject
-    UUIDUtil() {
+    public UUIDUtil() {
     }
 
     public List<UUID> extractUUIDs(byte[] scanResult) {


dariuszseweryn:
Hello
Why would you like to use UUIDUtil? It is in `internal` package and is not intended for public accesibility

marciogranzotto:
I would like to use it's `extractUUIDs` function to get the UUIDs of a scanResult object.

lukevr:
Hi,
I am playing with RxAndroidBle 1.2.1 and have got an issue with scanBleDevices(UUIDArray), seems if I have UUIDArray with several different UUID, scanning wont work. Well, at least no results for subscriber in this case. 
If I add just one UUID, or same UUID several time - rxBleScanResult is present, and everything is ok.

Found similar issue on stackoverflow:
http://stackoverflow.com/questions/40447272/rxandroidble-the-method-scanbledevicesuuid-filters-is-not-support-for-two-t

Custom filter and UUIDUtil are suggested to be used. And, yes UUIDUtil is not public in 1.2.1.

What is the best way to solve the issue with scanning a set of different UUID?

Thanks in advance,
Victor.

dariuszseweryn:
@marciogranzotto I will think today how to properly address the issue. The `UUIDUtil` is in internal package and nothing from there should be used outside of the library. I will probably add some `UUIDHelper` and add it to the `helper` package.
@lukevr The answer is provided in the SO post you have referenced. A custom filter.

dariuszseweryn:
I have just released `1.2.2` with this change and `AdvertisedServiceUUIDExtractor` helper which you should use instead of an internal `UUIDUtil`.

It should be visible shortly.

Best Regards

uKL:
Dear Contributor,

Issue #231, opened 2017-06-23
`Observable` emitted from `RxBleConnection.setupNotification()`/`RxBleConnection.setupIndication()` does not complete


dariuszseweryn:
### Summary
Currently when the user will unsubscribe from `.setupNotification()`/`.setupIndication()` the notification/indication is teardown but the (possibly) emitted `Observable<byte[]>` does not complete.
Similarly when the `RxBleConnection` terminates the emitted `Observable<byte[]>` does not throw an error.
### Library version
`1.3.1`
#### Steps to reproduce actual result
<br />1. Connect to a device
<br />2. Setup notification (or indication)
<br />3. Wait for it being setup
<br />4. Get emitted `Observable<byte[]>`
<br />5. Un-subscribe from the `.setup*()`
<br />6. Observe that emitted `Observable<byte[]>` has not completed
#### Minimum code snippet reproducing the issue
```
rxBleDevice.establishConnection(false)
                .flatMap(connection -> connection.setupNotification(characteristicUuid)
                        .takeUntil(Observable.timer(15, TimeUnit.SECONDS))
                        .flatMap(observable -> observable)
                        .doOnCompleted(() -> Log.d("SHOULD", "COMPLETE"))
                )
                .subscribe();
```
#### Actual result
`SHOULD COMPLETE` is not logged after 15 seconds from the notification setup.
#### Expected result
`SHOULD COMPLETE` should be logged after 15 seconds from the notification setup.

RobLewis:
Haven't thought this through, but will this fix require any changes to `ConnectionSharingAdapter`?

dariuszseweryn:
@RobLewis No.

PR #232, opened 2017-06-23
Fix notify/indicate observable termination

dariuszseweryn:
Related to #231 
Until now the `Observable<byte[]>` emitted from `RxBleConnection.setupNotification()`/`RxBleConnection.setupIndication()` did not complete when the `RxBleConnection.setupNotification()`/`RxBleConnection.setupIndication()` was unsubscribed and notification/indication was teardown. Additionally the `Observable<byte[]>` did not propagate errors when the `RxBleConnection` has disconnected. This behaviour is fixed with this Pull Request.

### DIFF
--- a/rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/NotificationAndIndicationManagerTest.groovy
+++ b/rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/NotificationAndIndicationManagerTest.groovy
@@ -354,7 +354,7 @@ class NotificationAndIndicationManagerTest extends RoboSpecification {
     }
 
     @Unroll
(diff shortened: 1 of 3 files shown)


rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/connection/NotificationAndIndicationManagerTest.groovy
```diff
@@ -349,6 +353,46 @@ class NotificationAndIndicationManagerTest extends RoboSpecification {
         ].combinations()
     }
 
+    @Unroll
+    def "should complete the emitted Observable<byte> when un-subscribed"() {
```
uKL: `-` is not required here.

PR #234, opened 2017-06-27
Added pre-scan verification for excessive scan

dariuszseweryn:
https://github.com/Polidea/RxAndroidBle/issues/227
Android API 24 (7.0) introduced an undocomented “feature” that aims on applications that try to start too many scans (5+) in a short time window (30 seconds) by silently (log error only) not starting the scan. More info available here: https://blog.classycode.com/undocumented-android-7-ble-behavior-changes-d1a9bd87d983


dariuszseweryn:
There is an alternative on how to check for this error but it would need to parse the log on the device. Probably not the most performant but an interesting idea though.

passsy:
It's not possible to parse Logcat from the app

dariuszseweryn:
It is possible but only to parse your own app logcat. Unfortunately the log about excessive scans comes from the system and is not reachable.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/scan/ExcessiveScanCheckerApi24.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    private static final int SCANS_LENGTH = 5;
+
+    private static final long EXCESSIVE_SCANNING_PERIOD = 30 * 1000;
+
+    private final long[] previousChecks = new long[SCANS_LENGTH];
+
+    @Nullable
+    @Override
+    public Date suggestDateToRetry() {
```
uKL: It's not obvious that the method returning a suggestion is also modifying the state of the checker. I think that this method should idempotent and there should be a method to alter the state.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/scan/ExcessiveScanCheckerApi24.java
```diff
@@ -0,0 +1,45 @@
...
+        final long oldestCheckTimestamp = previousChecks[oldestCheckTimestampIndex];
+        final long currentCheckTimestamp = System.currentTimeMillis();
+        if (currentCheckTimestamp - oldestCheckTimestamp < EXCESSIVE_SCANNING_PERIOD) {
+            return new Date(oldestCheckTimestamp + EXCESSIVE_SCANNING_PERIOD);
+        }
+        previousChecks[oldestCheckTimestampIndex] = currentCheckTimestamp;
+        return null;
+    }
+
+    private int getOldestCheckTimestampIndex() {
```
uKL: Something like a fixed size stack seems to be cool https://stackoverflow.com/a/16206356

Issue #165, opened 2017-04-06
Access to low level bluetooth callback in custom operations


uKL:
In order to improve throughput more, I'd like to have more direct access (skipping Rx overhead) to BluetoothGattCallback.

Let's discuss how.

dariuszseweryn:
First idea - add a callback listener for every possible callback of `BluetoothGattCallback`

uKL:
This was my first idea, with a few assumptions:

1. Only one callback listener can be plugged in at time
1. It sticks to a custom operation paradigm
1. The callback listener is disposed (null) at the end of the custom operation automatically to prevent leaks.

What do you think?

dariuszseweryn:
Why `1. Only one callback listener can be plugged in at time`?

uKL:
To simplify the whole thing and avoid list iteration. In the scope of a single custom operation, I think it will be enough. What's your opinion?

dariuszseweryn:
Ah, one at a time in this meaning. That's reasonable. I first thought of a Callback Listener per callback type (one callback for `onCharacteristicWrite()` other for `onCharacteristicChanged()` and so on) and that one could not set both at the same time.

uKL:
Nope, I'm thinking about some abstract class implementing **BluetoothGattCallback** providing default, empty methods as templates. The user would be able to implement whichever callback required and only that one.

dariuszseweryn:
Is available in `1.4.0-SNAPSHOT`

RobLewis:
This sounds interesting. Is there any documentation of how it fits into the `RxAndroidBle` overall scheme? Usage example?

uKL:
Hey Rob,
all you need can be found in `RxBleGattCallback#setNativeCallback` Javadoc. The BluetoothGattCallback interface is documented in Android’s documentation.

PR #237, opened 2017-06-29
[#165] Added support for native BluetoothGattCallback usage in custom…

uKL:
Added support for native BluetoothGattCallback usage in custom operations.

Really silly performance test:
- based on JUnit
- invoke onCharacteristicRead like the system would do 1000000 times
- assert on number of read events received from RxJava based implementation and the native one
- test does not include load and delay introduces by the ExecutorService, wrapping emissions through the RxJava API

RxJava API @ 1000000 calls => 7905ms (0.008ms/call)
Native API @ 1000000 calls => 2076ms (0.002ms/call)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Change Log
 ==========
 
+Version 1.4.0-SNAPSHOT
+* Added native callback usage support in custom operations. You may consider this API if your implementation is performance critical. (https://github.com/Polidea/RxAndroidBle/issues/165)
+
 Version 1.3.2
 * Fixed completing the `Observable<byte[]>` emitted by `RxBleConnection.setupNotification()`/`RxBleConnection.setupIndication()` when unsubscribed (https://github.com/Polidea/RxAndroidBle/issues/231)
 
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/NativeCallbackDispatcher.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/NativeCallbackDispatcher.java
@@ -1,0 +1,85 @@
+package com.polidea.rxandroidble.internal.connection;
+
+import android.annotation.TargetApi;
+import android.bluetooth.BluetoothGatt;
+import android.bluetooth.BluetoothGattCallback;
+import android.bluetooth.BluetoothGattCharacteristic;
+import android.bluetooth.BluetoothGattDescriptor;
+import android.os.Build;
+
+import javax.inject.Inject;
+
+class NativeCallbackDispatcher {
+
+    private BluetoothGattCallback nativeCallback;
+
+    @Inject
+    NativeCallbackDispatcher() {
+
+    }
+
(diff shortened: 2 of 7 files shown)


dariuszseweryn:
Looks good for me.

6bangs:
Cool! Can this be used for a long-running task concurrently with other usages `RxBleConnection` (writing and reading characteristics, etc)?

uKL:
@6bangs no, it's not recommended. When the operation is completed we are releasing the radio lock. Having multiple bluetooth operations not synchronized may cause abnormal behaviors.

Anyway, the native callback is automatically released when the operation is finished. Is there any reason you wanted to use it that way?

6bangs:
Yep. We deliver OTA updates transparently to our users while the device is otherwise in use (reading heart rate, etc). It works great currently, but it takes 4-6 hours to complete.

Maybe we could use this if we opened a second `RxBleConnection` to the device specifically for OTA. Is there any reason that wouldn't work?

dariuszseweryn:
Hello @6bangs 
Could you explain in a bit more detail how do you achieve that? You want to have a constant stream of Heart Rate with a fixed interval - do you perform a read operation once every second (or a different amount of time) for that or do you use notifications? The library is implementing a priority queue but it is not available through the API. It would be possible to add for instance a low priority characteristic write to the next API.
Again if you would describe your use-case in greater detail - I will be able to think about how to make it possible.

6bangs:
Of course. Our device uses characteristics not defined in the BLE spec for both OTA and heart rate measurement (among other things). We set up notifications for heart rate measurement via the usual `setupNotification` method of `RxBleConnection`. Most traffic happens via notifications, but during OTA we do a lot of writing until the OTA process is complete. We also write occasionally to update the state of the device (set the heart rate monitoring interval, etc).

We can perform all of these operations concurrently, and it's working really well. OTA just takes forever due to RxJava overhead.

With regard to `BleAlreadyConnectedException`, is that a limitation of `RxAndroidBle` or the platform? I haven't actually tried connecting multiple `BluetoothGatt`s to the same device with the same app, but I assumed it must be possible since multiple apps can connect to the same device at the same time.

RobLewis:
@6bangs Are you using the `ConnectionSharingAdapter` to multiplex your various reads and writes with a single RxBleConnection? I recently switched to this from my somewhat lame home-brewed solution and it seems to be working well. More [here](https://github.com/Polidea/RxAndroidBle/issues/185).

6bangs:
Yep, exactly. Sorry, I forgot that detail; wrote that code almost a year ago. If I recall, I found it in an `RxAndroidBle` example. I bet it's something that's useful enough it should probably be called out in the `README.md`.

dariuszseweryn:
@6bangs The `BleAlreadyConnectedException` is a conscious decision to prevent people from interfering with their own interactions with the connection from different places in the code. This decision was taken because it is impossible to create BLE connections in the same stateless manner as HTTP.

It is possible to achieve what you want but you would need to put the OTA and updating the state of the peripheral in the same `RxBleRadioOperationCustom` and manage writes all by yourself.

6bangs:
Ah, cool. Will look into that.

PR #238, opened 2017-06-29
Adjusted `RxBleDevice.observeConnectionStateChanges()` behaviour

dariuszseweryn:
https://github.com/Polidea/RxAndroidBle/issues/50
Previously `RxBleDevice.observeConnectionStateChanges()` was emitting immediately the initial state. The state was derived from user’s actions (i.e. `CONNECTING` was emitted right after user has subscribed to `RxBleDevice.establishConnection()`) rather than state in which the `BluetoothGatt` was. This behaviour has been changed to best reflect the state of `BluetoothGatt` at any given time.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationConnect.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationConnect.java
@@ -29,6 +29,7 @@ import rx.functions.Func0;
 import rx.functions.Func1;
 
 import static com.polidea.rxandroidble.RxBleConnection.RxBleConnectionState.CONNECTED;
+import static com.polidea.rxandroidble.RxBleConnection.RxBleConnectionState.CONNECTING;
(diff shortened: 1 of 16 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/RxBleRadioOperationConnect.java
```diff
@@ -48,12 +50,14 @@ public Builder(
                 BleConnectionCompat connectionCompat,
                 RxBleGattCallback rxBleGattCallback,
                 @Named(CONNECT_TIMEOUT) TimeoutConfiguration connectionTimeout,
-                BluetoothGattProvider bluetoothGattProvider) {
+                BluetoothGattProvider bluetoothGattProvider,
+                Action1<RxBleConnection.RxBleConnectionState> connectionStateChangedAction) {
```
uKL: It's not really declarative that we inject `Action1<RxBleConnection.RxBleConnectionState>` here. Can we have it wrapped in a custom interface saying that this is a connection state dispatcher? 

Looking from a test/user perspective it wouldn't say much about this parameter.

PR #239, opened 2017-06-30
Proposition: new `Connection` API (replace `RxBleConnection`)

dariuszseweryn:
It aims to provide a more concise API that utilises newer `RxJava 1` API like `Single` and `Completable` as well as a clean extension point for future functionality “knobs”.

This Pull Request is made to ignite a discussion about the future API.


RobLewis:
This was mentioned as being part of the 1.5.0 release. Is that still the plan?

uKL:
I don't think so, I think we'll postpone it. It looks like there will be 1.5.0, then RxJava2, then new API.

Am I right @dariuszseweryn ?

dariuszseweryn:
This will not be included in `1.5.0`. We currently have limited capacity and decided that support for `RxJava 2` is of greater importance. The new API will get included once it is ready... Hard to say when.

PR #240, opened 2017-07-03
Operation viability check

mikolak:
This PR is is connected to https://github.com/Polidea/RxAndroidBle/issues/224

I have implemented a mechanism for checking whether the requested characteristic operation can succeed. The check is based upon the properties of the chosen characteristic. End user can opt-out of the 'hard' way of checking (ie. throwing an exception) using new connection setup API.


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -195,6 +204,10 @@ public int getMtu() {
     @Override
     public Observable<Observable<byte[]>> setupIndication(@NonNull BluetoothGattCharacteristic characteristic,
                                                           @NonNull NotificationSetupMode setupMode) {
+        illegalOperationChecker.checkAnyPropertyMatches(
```
dariuszseweryn: I am actually thinking wether or not it would be better to have something like:
```
if (illegalOperationChecker.checkAnyPropertyMatches()) {
  return ... // normal stuff
} else {
  return Observable.error(new IllegalOperationException(/* construct */));
}
```
It would be then consistent no matter how the user would call the function `RxBleConnection.setupIndication(UUID)` / `RxBleConnection.setupIndication(BluetoothGattCharacteristic)`. Because now calling `RxBleConnection.setupIndication(BluetoothGattCharacteristic)` will just throw.

It would need a different abstraction though. @uKL any thoughts?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/RxBleDeviceImpl.java
```diff
@@ -53,12 +54,20 @@
 
     @Override
     public Observable<RxBleConnection> establishConnection(final boolean autoConnect) {
+        ConnectionSetup options = (new ConnectionSetup.Builder())
+                .setAutoConnect(autoConnect)
+                .build();
```
dariuszseweryn: This should also have `.setSuppressOperationCheck(false)` to match the previous behaviour.

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
```diff
@@ -0,0 +1,42 @@
...
+
+        private boolean autoConnect = false;
+        private boolean supressOperationCheck = false;
+
+        public Builder setAutoConnect(boolean autoConnect) {
+            this.autoConnect = autoConnect;
+            return this;
+        }
+
+        public Builder setSupressOperationCheck(boolean suppressOperationCheck) {
```
dariuszseweryn: `setSuppressIllegalOperationCheck()`? Now it is not too informative in my opinion.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
```diff
@@ -55,9 +56,11 @@ public ConnectorImpl(
                     return Observable.error(new BleDisconnectedException(bluetoothDevice.getAddress()));
                 }
 
-                final ConnectionComponent connectionComponent = connectionComponentBuilder.build();
+                final ConnectionComponent connectionComponent = connectionComponentBuilder
+                        .connectionModule(new ConnectionModule(options.suppressOperationCheck))
+                        .build();
                 RxBleRadioOperationConnect operationConnect = connectionComponent.connectOperationBuilder()
```
dariuszseweryn: It should be now possible to get rid of the `RxBleRadioOperationConnect.Builder` as all dependencies could be passed through `ConnectionComponent.Builder`.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleDevice.java
```diff
@@ -81,9 +81,38 @@
...
+     * Establishes connection with a given BLE device. {@link RxBleConnection} is a handle, used to process BLE operations with a connected
+     * device.
+     * <p>
+     * The connection is automatically disconnected (and released) when resulting Observable is unsubscribed.
+     * On the other hand when the connections is interrupted by the device or the system, the Observable will be unsubscribed as well
+     * following BleDisconnectedException or BleGattException emission.
+     * <p>
+     * During the disconnect process the library automatically handles order and requirement of device disconnect and gatt close operations.
+     * <p>
+     * Autoconnect concept may be misleading at first glance. In cases when the BLE device is available and it is advertising constantly you
```
dariuszseweryn: This part about `autoConnect` should go to `ConnectionSetup` javadoc as it will be the place to set.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
mikolak: I have modified IllegalOperationChecker.checkAnyPropertyMatches() to return Completable.fromCallable() and will upload it shortly.

uKL: I agree that there should be a model separation. The method should check whether preconditions are correct and the result should be described by the caller.

uKL: Moreover, the point where the check happens is inconsistent. Sometimes it happens even before the operation is enqueued and sometimes not (when it's required to perform discovery in order to get the characteristic object)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/IllegalOperationChecker.java
```diff
@@ -0,0 +1,135 @@
...
+     * the supplied bitmask.
+     * @param characteristic a {@link BluetoothGattCharacteristic} the operation is done on
+     * @param neededProperties properties required for the operation to be successfully completed
+     */
+    public void checkAnyPropertyMatches(BluetoothGattCharacteristic characteristic,
+                                        @BluetoothGattCharacteristicProperty int neededProperties) {
+        final int characteristicProperties = characteristic.getProperties();
+
+        if ((characteristicProperties & neededProperties) == 0) {
+            int[] possibleProperties = getPossibleProperties();
```
uKL: I guess it's more like `permited` or `allowed` than `possible`, right?

uKL: or even `supported` if I'm reading correctly from the context.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/IllegalOperationChecker.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Method for handling the potential non-match message
+     * @param message message describing the
+     */
+    protected abstract void handleMessage(String message);
+
+    @NonNull
+    private String propertiesIntToString(int property, int[] possibleProperties) {
```
uKL: More like an util. What do you think about extracting it?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/IllegalOperationChecker.java
```diff
@@ -0,0 +1,135 @@
...
+            );
+            handleMessage(message);
+        }
+    }
+
+    /**
+     * Method for handling the potential non-match message
+     * @param message message describing the
+     */
+    protected abstract void handleMessage(String message);
```
uKL: I'd prefer the checker to return the model with the report, rather than to pass it to a method that needs to be overridden. @dariuszseweryn what do you think? It's easier to test then.

dariuszseweryn: Currently we have a `IllegalOperationChecker` which is checking and handling the result. We could have `IllegalOperationChecker` that would work with `IllegalOperationHandler` which would either log or throw/emit an `IllegalOperationException`. The checker part would be always the same (could be put directly in `RxBleConnectionImpl` or injected as a separate object)

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
```diff
@@ -34,6 +34,10 @@ private ConnectionSetup(boolean autoConnect, boolean suppressOperationCheck) {
 
 
         /**
+         * Autoconnect concept may be misleading at first glance. In cases when the BLE device is available and it is advertising constantly
+         * you won't need to use autoconnect. Use autoconnect for connections where the BLE device is not advertising at
+         * the moment of #establishConnection call.
```
dariuszseweryn: `{@link com.polidea.rxandroidble.RxBleConnection#establishConnection(ConnectionSetup)}` instead of `#establishConnection`

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
```diff
@@ -1,11 +1,24 @@
...
+ * characteristic with only read property.
  */
 public class BleIllegalOperationException extends RuntimeException {
 
-    public BleIllegalOperationException(String message) {
+    public final UUID characteristicUUID;
+    public final int supportedProperties;
+    public final int neededProperties;
+
+    public BleIllegalOperationException(String message,
```
dariuszseweryn: Add `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` and a javadoc that will tell people not to instantiate this class. Maybe it is worth to expose getters instead of `public final`s — just a thought.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectionModule.java
```diff
@@ -34,21 +34,23 @@ public ConnectionModule(ConnectionSetup connectionSetup) {
     @ConnectionScope
     IllegalOperationChecker provideIllegalOperationChecker() {
         if (suppressOperationCheck) {
-            return new LoggingIllegalOperationChecker(BluetoothGattCharacteristic.PROPERTY_BROADCAST,
+            return new IllegalOperationChecker(BluetoothGattCharacteristic.PROPERTY_BROADCAST,
```
dariuszseweryn: This may be now optimised. `if else` only picks a `MismatchDataHandler`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/IllegalOperationChecker.java
```diff
@@ -79,18 +83,30 @@ public Object call() throws Exception {
                             propertiesIntToString(neededProperties, possibleProperties),
                             neededProperties
                     );
-                    handleMessage(message);
+                    MismatchData mismatchData = new MismatchData(message,
```
dariuszseweryn: I think that having a `MismatchData` here is superfluous and it only adds an allocation. The data could be passed directly to the `MismatchDataHandler`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/MismatchDataHandler.java
```diff
@@ -0,0 +1,10 @@
+package com.polidea.rxandroidble.internal.util;
+
+
+/**
+ * Handler for {@link com.polidea.rxandroidble.internal.util.IllegalOperationChecker.MismatchData} returned by
+ * {@link IllegalOperationChecker}.
+ */
+public interface MismatchDataHandler {
```
dariuszseweryn: Why not just a `IllegalOperationHandler`? And it can be put just in `internal.connection` package. Does not look like an util that is used all over the code.

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
mikolak: Well, getters will increase the method count and all of these fields are immutable anyhow. Can be done, though I'm not sure if it should be.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/util/IllegalOperationChecker.java
mikolak: I'm not sure. If it's supposed to be testable, then it would need to be instantiated. That would lead to increasing method count and I'm not that convinced that it would be useful anywhere else.

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
```diff
@@ -0,0 +1,69 @@
...
+         */
+        public Builder setAutoConnect(boolean autoConnect) {
+            this.autoConnect = autoConnect;
+            return this;
+        }
+
+        /**
+         * @param suppressOperationCheck Flag describing the method of operation viability checking. If set to false,
+         *                               a {@link com.polidea.rxandroidble.exceptions.BleIllegalOperationException} will be
+         *                               thrown everytime properties of the characteristic don't match those required by the operation.
```
dariuszseweryn: `thrown` -> `emitted`

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
```diff
@@ -0,0 +1,27 @@
...
+
+import android.support.annotation.RestrictTo;
+
+import java.util.UUID;
+
+/**
+ * This exception is thrown when a non-supported operation has been requested upon a characteristic, eg. write operation on a
+ * characteristic with only read property.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
dariuszseweryn: the annotation should be on the constructor only. the class is meant to be used in the API

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectionComponent.java
```diff
@@ -7,17 +7,25 @@
...
+
     @Subcomponent.Builder
     interface Builder {
 
+        Builder connectionModule(ConnectionModule connectionModule);
+
         ConnectionComponent build();
     }
 
     @ConnectionScope
```
dariuszseweryn: Maybe it is possible to get rid of the `@ConnectionScope` here as the operation is used only once and it is not needed to be cached. It is worth checking.

dariuszseweryn: Maybe it is possible to get rid of caching for other items as well. please check method counts / fields.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectionModuleBinder.java
```diff
@@ -0,0 +1,41 @@
...
+import dagger.Binds;
+import javax.inject.Named;
+
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+abstract public class ConnectionModuleBinder {
+
+    static final String GATT_WRITE_MTU_OVERHEAD = "GATT_WRITE_MTU_OVERHEAD";
```
dariuszseweryn: Maybe this should also be with other names in the `ConnectionComponent`?

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
dariuszseweryn: True though here we have fields in the API and we cannot change that later on if we would decide to pass `BluetoothGattCharacteristic` in the constructor. But it is just an opinion. @uKL thoughts?

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
```diff
@@ -0,0 +1,69 @@
...
+        private boolean autoConnect = false;
+        private boolean suppressOperationCheck = false;
+
+
+        /**
+         * Autoconnect concept may be misleading at first glance. In cases when the BLE device is available and it is advertising constantly
+         * you won't need to use autoconnect. Use autoconnect for connections where the BLE device is not advertising at
+         * the moment of {@link RxBleDevice#establishConnection(ConnectionSetup)} call.
+         *
+         * @param autoConnect Marker related with
```
dariuszseweryn: `related to` or `connected with` ;) `Flag` maybe?

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
```diff
@@ -0,0 +1,69 @@
...
+         *                    {@link android.bluetooth.BluetoothDevice#connectGatt(Context, boolean, BluetoothGattCallback)} autoConnect
+         *                    flag. In case of auto connect is enabled the observable will wait with the emission of RxBleConnection.
+         *                    Without auto connect flag set to true the connection will fail
+         *                    with {@link com.polidea.rxandroidble.exceptions.BleGattException} if the device is not in range.
```
dariuszseweryn: `after around 30 seconds timeout`

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
mikolak: My fault for copy-paste without proof reading, not my grammar mistake. ;) Should I correct it elsewhere as well?

rxandroidble/src/main/java/com/polidea/rxandroidble/ConnectionSetup.java
mikolak: Should I correct it elsewhere as well?

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
uKL: Well, the first thought is "Why do I need an int?! It's ugly.". If we'd have an API to read properties more easily.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectionComponent.java
uKL: I think further optimization is out of scope for this pull request. We should have a separate PR for optimizations if you think they are necessary.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/ThrowingIllegalOperationHandlerTest.groovy
```diff
@@ -0,0 +1,27 @@
...
+    int neededProperties = 0
+
+
+    void setup() {
+        objectUnderTest = new ThrowingIllegalOperationHandler()
+    }
+
+    def "should throw BleIllegalOperationException if no property matches"() {
+        when:
+        objectUnderTest.handleMismatchData(message, mockUuid, supportedProperties, neededProperties)
```
uKL: the name suggests a mock for `mockUuid` but it's null.

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/IllegalOperationCheckerTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+class IllegalOperationCheckerTest extends Specification {
+
+    IllegalOperationChecker objectUnderTest
+    BluetoothGattCharacteristic bluetoothGattCharacteristic
+    TestSubscriber subscriber
+    IllegalOperationHandler mockHandler
+
+    void setup() {
+        mockHandler = Mock IllegalOperationHandler
+        subscriber = new  TestSubscriber()
```
uKL: too many spaces

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/IllegalOperationCheckerTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+                BluetoothGattCharacteristic.PROPERTY_SIGNED_WRITE,
+                mockHandler
+        )
+
+        bluetoothGattCharacteristic = Mock BluetoothGattCharacteristic
+    }
+
+    def "should handle message if no property matches"() {
+        given:
+        bluetoothGattCharacteristic.getProperties() >> 0
```
uKL: What stands for `0`?

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/IllegalOperationCheckerTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+
+        when:
+        objectUnderTest.checkAnyPropertyMatches(bluetoothGattCharacteristic, BluetoothGattCharacteristic.PROPERTY_READ)
```
uKL: Is `subscriber` used anyhow?

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/IllegalOperationCheckerTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+        then:
+        1 * mockHandler.handleMismatchData(_, _, _, _)
+    }
+
+    def "should not handle message if the only needed property matches"() {
+        given:
+        bluetoothGattCharacteristic.getProperties() >> BluetoothGattCharacteristic.PROPERTY_READ
+
+        when:
+        objectUnderTest.checkAnyPropertyMatches(bluetoothGattCharacteristic, BluetoothGattCharacteristic.PROPERTY_READ)
```
uKL: Is `subscriber` used anyhow?

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/util/IllegalOperationCheckerTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+        then:
+        0 * mockHandler.handleMismatchData(_, _, _, _)
+    }
+
+    def "should not handle message if at least one of the needed properties match"() {
+        given:
+        bluetoothGattCharacteristic.getProperties() >> BluetoothGattCharacteristic.PROPERTY_WRITE
+
+        when:
+        objectUnderTest.checkAnyPropertyMatches(bluetoothGattCharacteristic,
```
uKL: Is `subscriber` used anyhow?

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/RxBleDeviceTest.groovy
```diff
@@ -210,4 +336,49 @@ public class RxBleDeviceTest extends Specification {
...
+
+        EstablishConnectionTestSetup(EstablishConnectionCaller establishConnectionCaller, ConnectionSetup connectionSetup) {
+            this.establishConnectionCaller = establishConnectionCaller
+            this.connectionSetup = connectionSetup
+        }
+
+        Observable<RxBleConnection> establishConnection(RxBleDevice d) {
+            return establishConnectionCaller.connectionStartClosure.call(d, connectionSetup)
+        }
+
```
uKL: Remove empty line

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.concurrent.Callable;
+
+import rx.Completable;
+
+/**
+ * Class for checking whether the requested operation is legal on chosen characteristic.
+ */
+public class IllegalOperationChecker {
+
+    private int propertyBroadcast;
```
uKL: Please make attributes final, just in case.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
```diff
@@ -0,0 +1,142 @@
...
+                    );
+                    resultHandler.handleMismatchData(message, characteristic.getUuid(), characteristicProperties, neededProperties);
+                }
+                return null;
+            }
+        });
+    }
+
+    @NonNull
+    private String propertiesIntToString(int property, int[] possibleProperties) {
```
uKL: I still think generating a message is not in scope of the checker, it should be a separate util class.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
```diff
@@ -0,0 +1,142 @@
...
+        StringBuilder builder = new StringBuilder();
+        builder.append(" ");
```
uKL: Why is there a space at the beginning?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
```diff
@@ -0,0 +1,142 @@
...
+            return "SIGNED_WRITE";
+        } else if (property == propertyIndicate) {
+            return "INDICATE";
+        } else if (property == propertyBroadcast) {
+            return "BROADCAST";
+        } else if (property == propertyNotify) {
+            return "NOTIFY";
+        } else {
+            RxBleLog.e("Unknown property specified");
+            throw new IllegalArgumentException("Unknown property specified");
```
uKL: Is it documented that the method may throw IllegalArgumentException? It will be passed to the public API.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
mikolak: For prettyness. [ ] looks better than [] if there are no properties and [ READ WRITE ] looks better than [READ WRITE ]. I decided to avoid tracking whether any property has already been matched or not. I admit that I wrote specifically for the message which puts the string in square brackets, so for the purpose of extracting it to a separate parser, maybe the brackets should be added to this function?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
mikolak: True, I'll extract it.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
mikolak: Just removed the exception in a refactor. ;)

rxandroidble/src/main/java/com/polidea/rxandroidble/exceptions/BleIllegalOperationException.java
dariuszseweryn: Maybe we could annotate the ints with `@BluetoothGattCharacteristicProperty` to make it more precise what is this `int` related to. `BluetoothGattCharacteristicProperty` would then be put on the public API.

rxandroidble/src/main/java/com/polidea/rxandroidble/BluetoothGattCharacteristicProperty.java
```diff
@@ -0,0 +1,3 @@
+package com.polidea.rxandroidble;
+
+public @interface BluetoothGattCharacteristicProperty { }
```
dariuszseweryn: It still needs proper annotations?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationHandler.java
```diff
@@ -14,5 +16,5 @@
         this.messageCreator = messageCreator;
     }
 
-    public abstract void handleMismatchData(BluetoothGattCharacteristic characteristic, int neededProperties);
+    public abstract BleIllegalOperationException handleMismatchData(BluetoothGattCharacteristic characteristic, int neededProperties);
```
dariuszseweryn: mark it with `@Nullable`

rxandroidble/src/main/java/com/polidea/rxandroidble/BluetoothGattCharacteristicProperty.java
dariuszseweryn: ...aaaaaand javadoc ;)

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleDevice.java
```diff
@@ -31,11 +31,12 @@
 
     /**
      * @param context     Android's context.
-     * @param autoConnect Marker related with
+     * @param autoConnect Flag related to
```
uKL: In the `ConnectionSetup` it still says "marker"

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/IllegalOperationChecker.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * This method checks whether the supplied characteristic possesses properties supporting the requested kind of operation, specified by
+     * the supplied bitmask.
+     *
+     * @param characteristic   a {@link BluetoothGattCharacteristic} the operation is done on
+     * @param neededProperties properties required for the operation to be successfully completed
+     * @return {@link Completable} deferring execution of the check till subscription
+     * @throws BleIllegalOperationException if there was no match between supported and necessary properties of characteristic and check has
```
uKL: In fact it does not throw, it emits BleIllegalOperationException

uKL:
Dear @mikolak ,

PR #243, opened 2017-07-05
Fixed scan filtering by name for API <21.

dariuszseweryn:
As it was observed for API <21 the name is not always available in the `ScanRecord` but it is available in the `BluetoothDevice`. On API >=21 this issue is not visible.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/scan/ScanFilter.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/scan/ScanFilter.java
@@ -277,8 +277,12 @@ import java.util.UUID;
         }
 
         // Local name match.
-        if (mDeviceName != null && !mDeviceName.equals(scanRecord.getDeviceName())) {
-            return false;
+        if (mDeviceName != null) {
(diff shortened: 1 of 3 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/scan/ScanFilter.java
```diff
@@ -277,8 +277,12 @@ public boolean matches(RxBleInternalScanResult scanResult) {
         }
 
         // Local name match.
-        if (mDeviceName != null && !mDeviceName.equals(scanRecord.getDeviceName())) {
-            return false;
+        if (mDeviceName != null) {
+            final boolean advertisedDeviceNameMatches = mDeviceName.equals(scanRecord.getDeviceName());
+            final boolean bluetoothDeviceNameMatches = mDeviceName.equals(device.getName());
```
uKL: Small but pricy optimization. You can inline check from line 282 (and extract to a method) to avoid code execution when it's not required.

In this example the second check will be executed regardless the first one (even if deviceName matches).

PR #255, opened 2017-07-27
Thread calling the Native API changed from main to background.

dariuszseweryn:
Initially when working with the Android Bluetooth API (18 / Android 4.3) it was found that specific implementations (Samsung’s) cannot work correctly if called directly in callbacks from other calls. To mitigate the problem the next interaction was routed to Android main thread. Now when investigating the issue it turned out that it may be any other thread as long as it is not the callback thread.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/ClientComponent.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/ClientComponent.java
@@ -9,6 +9,7 @@ import android.content.pm.PackageManager;
 import android.location.LocationManager;
 import android.os.Build;
 import android.support.annotation.Nullable;
+import android.support.annotation.RestrictTo;
 import com.polidea.rxandroidble.helpers.LocationServicesOkObservable;
 import com.polidea.rxandroidble.internal.DeviceComponent;
 import com.polidea.rxandroidble.internal.scan.InternalToExternalScanResultConverter;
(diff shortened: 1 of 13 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/ClientComponent.java
```diff
@@ -119,28 +118,50 @@ static int provideDeviceSdk() {
...
-        static Scheduler provideMainThreadScheduler() {
-            return AndroidSchedulers.mainThread();
+        static ClientComponentFinalizer provideFinalizationCloseable(
+                @Named(NamedSchedulers.BLUETOOTH_INTERACTION) final ExecutorService interactionExecutorService,
+                @Named(NamedSchedulers.BLUETOOTH_CALLBACKS) final ExecutorService callbacksExecutorService
+        ) {
+            return new ClientComponentFinalizer() {
+                @Override
+                public void onFinalize() {
+                    interactionExecutorService.shutdown();
```
uKL: Actually this is incorrect. Dagger will provide you a new instance of the executor each time you ask for it. You seems to be closing an instance that wasn't even started and you're leaving started service as is. 

Provided executors should be scoped.

dariuszseweryn: From what I see the executors are scoped. And `.shutdown()` is executed in `RxBleClient.finalize()`. Or am I missing something?

dariuszseweryn:
Naturally this change will be described in the Changelog.

rxandroidble/src/main/java/com/polidea/rxandroidble/ClientComponent.java
uKL: You are correct, I missed the annotation `@ClientScope`

Issue #263, opened 2017-08-07
Tag getName() as Nullable in RxBleDevice


vpratfr:
When using Kotlin, bleDevice.name is a property which is assumed to be null-safe.

This is not the case as name can be null.

An annotation `@Nullable` should hint Kotlin that the generated name property can have null values and should be considered as `String?` instead of simply `String`

Noticed that when getting a NPE in my code:

`scanResult.bleDevice.name.startsWith("EXAMPLE-", true)` give an exception: 

```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter $receiver
    at kotlin.text.StringsKt__StringsJVMKt.startsWith(StringsJVM.kt)
```

uKL:
Thanks for letting us know! It will be added soon.

dariuszseweryn:
Should be available in `1.3.4-SNAPSHOT` shortly.

PR #268, opened 2017-08-21
Added @Nullable annotation to a bluetooth device name. 

uKL:
It may be undefined as it's an optional broadcast parameter.

This fixes issue #263.


rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleDevice.java
```diff
@@ -83,6 +84,7 @@
      *
      * @return The device name or null if device name is absent.
      */
+    @Nullable
```
piotrek1543: Shouldn't be `@Nullable` annotation also in `RxBleDeviceMock` class?

PR #272, opened 2017-08-27
Fix lint error

mapyo:
Thank you this library!
I fix the following lint error.

```
String.format string doesn't match the XML format string
```

```
../../src/main/java/com/polidea/rxandroidble/sample/example5_rssi_periodic/RssiPeriodicExampleActivity.java:54: 
Suspicious argument type for formatting argument #1 in read_rssi: conversion is s, 
received int (argument #2 in method call) (Did you mean formatting character d, 'o' or x?)
```


dariuszseweryn:
Thank you for spotting and the fix!

PR #276, opened 2017-09-01
Changed disconnection error reporting in RxBleGattCallback.

dariuszseweryn:


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
```diff
@@ -51,35 +52,33 @@ public Object call(BleGattException bleGattException) {
...
+                    @Override
+                    public Boolean call(Pair<BluetoothGatt, RxBleConnectionState> pair) {
+                        RxBleConnectionState rxBleConnectionState = pair.second;
+                        return rxBleConnectionState == RxBleConnectionState.DISCONNECTED
+                                || rxBleConnectionState == RxBleConnectionState.DISCONNECTING;
+                    }
+                }
+        );
+        final Observable<?> otherConnectionErrors = gattAndConnectionStateOutput.errorRelay.map(errorMapper);
+        this.disconnectedErrorObservable = Observable.error(new BleDisconnectedException(bluetoothDevice.getAddress()))
```
uKL: Can we have the address only? We don't need the BluetoothDevice.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
```diff
@@ -51,35 +52,33 @@ public Object call(BleGattException bleGattException) {
...
+                .delaySubscription(isDisconnectedOrDisconnecting)
+                .mergeWith(otherConnectionErrors)
+                .replay()
+                .autoConnect(0);
```
uKL: As a matter of fact, did we analyze if this observable is unsubscribed at some point of time? `autoConnect` is a risky operator.

dariuszseweryn: This should not be a problem as every connection related object is being garbage collected once the connection is no longer used

Issue #275, opened 2017-09-01
RxBle is not show disconnected when BluetoothGatt: android.os.DeadObjectException is thrown


hoanglm4:
### Summary
CONNECTED state forever when BluetoothGatt: android.os.DeadObjectException is thrown

My code in below:
```
   private static final UUID characteristicUuid = UUID.fromString("25A80001-6478-11E6-BDF4-AD3BCC77759F");

   @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_example4);
        ButterKnife.bind(this);
        bleDevice = SampleApplication.getRxBleClient(this).getBleDevice(macAddress);
        connectionObservable = prepareConnectionObservable();
        bleDevice.observeConnectionStateChanges().subscribe(state -> {
            Log.i(getClass().getSimpleName(), "" + state.toString());
        });
        connect();
    }

    private Observable<RxBleConnection> prepareConnectionObservable() {
        return bleDevice
                .establishConnection(false)
                .takeUntil(disconnectTriggerSubject)
                .compose(bindUntilEvent(PAUSE))
                .compose(new ConnectionSharingAdapter());
    }

    @OnClick(R.id.connect)
    public void connect() {
        connectionObservable.flatMap(RxBleConnection::discoverServices)
                .flatMap(rxBleDeviceServices -> rxBleDeviceServices.getCharacteristic(characteristicUuid))
                .observeOn(AndroidSchedulers.mainThread())
                .doOnSubscribe(() -> Log.i(getClass().getSimpleName(), "start connect "))
                .subscribe(
                        characteristic -> {
                            Log.i(getClass().getSimpleName(), "Hey, connection has been established!");
                        },
                        this::onConnectionFailure,
                        this::onConnectionFinished
                );
    }

   @OnClick(R.id.dissconnect)
    public void triggerDisconnect() {
        disconnectTriggerSubject.onNext(null);
    }

     @OnClick(R.id.read_rssi)
    public void readRSSI() {
        if (bleDevice.getConnectionState() != RxBleConnection.RxBleConnectionState.CONNECTED) {
            return;
        }
        connectionObservable.flatMap(rxBleConnection -> rxBleConnection.readRssi())
                .subscribe(rssi -> {
                    Log.i(getClass().getSimpleName(), "RSSI = " + rssi);
                }, throwable -> {
[...]

PR #278, opened 2017-09-05
Added Bluetooth Adapter state monitoring to connection (#275)

dariuszseweryn:
`DeadObjectException` is usually thrown when interacting with `BluetoothAdapter` or `BluetoothGatt` instance that was obtained before bluetooth being turned off. Prior to library version `1.3.0` all errors raised when interacting with `BluetoothGatt` or recieved by `BluetoothGattCallback` were closing the connection (were emitted by `RxBleDevice.establishConnection()`). After `1.3.0` only errors raised in `RxBleRadioOperationConnect` and recieved by `BluetoothGattCallback.onConnectionStateChange()` were closing the connection so it was possible that `DeadObjectException`s raised repetedly by subscribing to i.e. `RxBleConnection.readRssi()` would not close the connection. Added monitoring of BluetoothAdapter’s state to prevent `DeadObjectException`s and inform the user about the connection loss as soon as possible.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
@@ -1,0 +1,74 @@
+package com.polidea.rxandroidble.internal.connection;
+
+
+import com.jakewharton.rxrelay.BehaviorRelay;
+import com.polidea.rxandroidble.RxBleAdapterStateObservable;
+import com.polidea.rxandroidble.exceptions.BleDisconnectedException;
+import com.polidea.rxandroidble.exceptions.BleException;
+import com.polidea.rxandroidble.internal.DeviceModule;
+import javax.inject.Inject;
+import javax.inject.Named;
+import rx.Observable;
+import rx.functions.Func1;
+
+/**
+ * A class that is responsible for routing all potential sources of disconnection to an Observable that emits only errors.
+ */
+@ConnectionScope
+class DisconnectionRouter {
+
+    private final BehaviorRelay<BleException> disconnectionErrorRelay = BehaviorRelay.create();
+
+    private final Observable disconnectionErrorObservable;
+
+    @Inject
+    DisconnectionRouter(
+            @Named(DeviceModule.MAC_ADDRESS) final String macAddress,
+            Observable<RxBleAdapterStateObservable.BleAdapterState> adapterStateObservable
+    ) {
+        disconnectionErrorObservable = Observable.merge(
+                disconnectionErrorRelay
+                        .flatMap(new Func1<BleException, Observable<?>>() {
+                            @Override
+                            public Observable<?> call(BleException e) {
+                                return Observable.error(e);
+                            }
+                        }),
+                adapterStateObservable
+                        .filter(new Func1<RxBleAdapterStateObservable.BleAdapterState, Boolean>() {
+                            @Override
+                            public Boolean call(RxBleAdapterStateObservable.BleAdapterState bleAdapterState) {
+                                return !bleAdapterState.isUsable();
+                            }
+                        })
+                        .flatMap(new Func1<RxBleAdapterStateObservable.BleAdapterState, Observable<?>>() {
+                            @Override
+                            public Observable<?> call(RxBleAdapterStateObservable.BleAdapterState bleAdapterState) {
+                                return Observable.error(new BleDisconnectedException(macAddress)); // TODO: Introduce BleDisabledException?
+                            }
+                        })
+        )
+                .replay()
+                .autoConnect(0);
+    }
+
+    /**
+     * Method to be called whenever a connection braking exception happens. It will be routed to {@link #asObservable()}.
+     *
+     * @param bleException the exception that happened
+     */
+    void route(BleException bleException) {
+        disconnectionErrorRelay.call(bleException);
+    }
+
+    /**
+     * Function returning an Observable that will only throw error in case of a disconnection
+     *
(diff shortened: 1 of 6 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
```diff
@@ -0,0 +1,74 @@
...
+                            }
+                        })
+                        .flatMap(new Func1<RxBleAdapterStateObservable.BleAdapterState, Observable<?>>() {
+                            @Override
+                            public Observable<?> call(RxBleAdapterStateObservable.BleAdapterState bleAdapterState) {
+                                return Observable.error(new BleDisconnectedException(macAddress)); // TODO: Introduce BleDisabledException?
+                            }
+                        })
+        )
+                .replay()
```
uKL: Do we need replay & autoConnect at the observable here or just when listening for adapter state?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
```diff
@@ -88,8 +69,17 @@ public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState
             super.onConnectionStateChange(gatt, status, newState);
             bluetoothGattProvider.updateBluetoothGatt(gatt);
 
-            propagateErrorIfOccurred(gattAndConnectionStateOutput, gatt, status, BleGattOperationType.CONNECTION_STATE);
-            gattAndConnectionStateOutput.valueRelay.call(new Pair<>(gatt, mapConnectionStateToRxBleConnectionStatus(newState)));
+            if (isDisconnectedOrDisconnecting(newState)) {
+                disconnectionRouter.route(new BleDisconnectedException(gatt.getDevice().getAddress()));
```
uKL: Is there any reason to accept `BleException` here? I can see that only two options are possible here. Maybe there should be a method to report:

- BleDisconnected
- BleGattException

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
dariuszseweryn: In my opinion yes as we would like to have it repeated to anyone who will subscribe later

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
dariuszseweryn: One for each?
Do you think it would be good to add another `.route()` methods if different exceptions will be applicable?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
uKL: Isn't a `BehaviorRelay` enough at the source?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleGattCallback.java
uKL: `BleException` is a general purpose exception. Form interface point of view it's not consistent with the class role - to route disconnections. I suggest names like "on ble disconnected", "handle gatt excption".

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
dariuszseweryn: Actually `disconnectionErrorRelay` should be a `PublishRelay`.
With this `.replay().autoConnect(0)` we are sure that both the `adapterStateObservable()` and `disconnectionErrorRelay` will get unsubscribed when any error/emission will happen

uKL: Ok, agreed.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionConnectorImpl.java
```diff
@@ -51,71 +35,31 @@ public RxBleConnectionConnectorImpl(
...
-                            @Override
-                            public Observable<RxBleConnection> call(final BluetoothGatt bluetoothGatt) {
-                                return Observable.merge(
-                                        justOnNext(connectionComponent.rxBleConnection()),
-                                        connectionComponent.gattCallback().<RxBleConnection>observeDisconnect()
-                                );
-                            }
-                        })
-                        .doOnUnsubscribe(disconnect(connectionComponent.disconnectOperation()));
+                return justOnNext(connectionComponent.rxBleConnection())
```
uKL: This is equivalent (i think better readable)

```
rxBleRadio.queue(operationConnect)
                        .map(new Func1<BluetoothGatt, RxBleConnection>() {
                            @Override
                            public RxBleConnection call(BluetoothGatt bluetoothGatt) {
                                return connectionComponent.rxBleConnection();
                            }
                        })
```

dariuszseweryn: I have extracted variables so maybe it will be better to read.
I like that there is no need to create additional `Func1<BluetoothGatt, RxBleConnection>` — seems that `.delaySubscription()` is a perfect match

PR #294, opened 2017-10-16
Fix MTU not updated when changed by the peripheral. (closes#293)

dariuszseweryn:
Previously the `RxBleConnection.currentMtu` was updated only when the MTU was updated by the central. This prevented Long Writes from utilising the increased Maximum Transfer Unit or the used from having a valid current MTU value—this made the utilisation of available bandwidth suboptimal.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
@@ -47,7 +47,7 @@ public class ConnectorImpl implements Connector {
                 });
                 final Observable<BluetoothGatt> connectedObservable = clientOperationQueue.queue(connectionComponent.connectOperation());
                 final Observable<RxBleConnection> disconnectedErrorObservable = connectionComponent.gattCallback().observeDisconnect();
-                final Set<ConnectionSubscriptionWatcher> connSubWatchers = connectionComponent.connectionSubscriptionAwares();
+                final Set<ConnectionSubscriptionWatcher> connSubWatchers = connectionComponent.connectionSubscriptionWatchers();
 
                 return Observable.merge(
                         newConnectionObservable.delaySubscription(connectedObservable),
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
@@ -6,10 +6,12 @@ import com.polidea.rxandroidble.RxBleConnection;
 import com.polidea.rxandroidble.internal.ConnectionSetup;
 import com.polidea.rxandroidble.internal.serialization.ClientOperationQueue;
 
+import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
 
 import rx.Observable;
+import rx.functions.Action0;
 import rx.functions.Func0;
 
 public class ConnectorImpl implements Connector {
@@ -45,14 +47,28 @@ public class ConnectorImpl implements Connector {
                 });
                 final Observable<BluetoothGatt> connectedObservable = clientOperationQueue.queue(connectionComponent.connectOperation());
                 final Observable<RxBleConnection> disconnectedErrorObservable = connectionComponent.gattCallback().observeDisconnect();
-                final DisconnectAction disconnect = connectionComponent.disconnectAction();
+                final Set<ConnectionSubscriptionWatcher> connSubWatchers = connectionComponent.connectionSubscriptionAwares();
 
                 return Observable.merge(
                         newConnectionObservable.delaySubscription(connectedObservable),
-                        disconnectedErrorObservable,
-                        connectionComponent.mtuWatcherCompletable().<RxBleConnection>toObservable()
+                        disconnectedErrorObservable
                 )
-                        .doOnUnsubscribe(disconnect);
+                        .doOnSubscribe(new Action0() {
+                            @Override
+                            public void call() {
+                                for (ConnectionSubscriptionWatcher csa : connSubWatchers) {
+                                    csa.onConnectionSubscribed();
+                                }
+                            }
+                        })
+                        .doOnUnsubscribe(new Action0() {
+                            @Override
+                            public void call() {
+                                for (ConnectionSubscriptionWatcher csa : connSubWatchers) {
+                                    csa.onConnectionUnsubscribed();
+                                }
+                            }
+                        });
             }
         });
     }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
@@ -47,9 +47,11 @@ public class ConnectorImpl implements Connector {
                 final Observable<RxBleConnection> disconnectedErrorObservable = connectionComponent.gattCallback().observeDisconnect();
                 final DisconnectAction disconnect = connectionComponent.disconnectAction();
 
-                return newConnectionObservable
-                        .delaySubscription(connectedObservable)
-                        .mergeWith(disconnectedErrorObservable)
+                return Observable.merge(
+                        newConnectionObservable.delaySubscription(connectedObservable),
+                        disconnectedErrorObservable,
+                        connectionComponent.mtuWatcherCompletable().<RxBleConnection>toObservable()
+                )
                         .doOnUnsubscribe(disconnect);
             }
         });
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
@@ -1,45 +1,47 @@
 package com.polidea.rxandroidble.internal.connection;
 
 
-import java.util.concurrent.atomic.AtomicInteger;
 import javax.inject.Inject;
 import javax.inject.Named;
-import rx.Completable;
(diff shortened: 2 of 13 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -83,6 +83,21 @@ public RxBleConnectionImpl(
         this.longWriteOperationBuilderProvider = longWriteOperationBuilderProvider;
         this.callbackScheduler = callbackScheduler;
         this.illegalOperationChecker = illegalOperationChecker;
+
+        // unsubscription is not needed as all objects are on @ConnectionScope and will get GCed together
```
uKL: The assumption on object lifecycle from within the object is generally not a good pattern, even if in this case it's true.

dariuszseweryn: Agreed

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -83,6 +83,21 @@ public RxBleConnectionImpl(
...
+        gattCallback.getOnMtuChanged()
```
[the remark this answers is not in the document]
dariuszseweryn: True

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/RxBleConnectionImpl.java
```diff
@@ -83,6 +83,21 @@ public RxBleConnectionImpl(
...
+                .retryWhen(new Func1<Observable<? extends Throwable>, Observable<?>>() {
```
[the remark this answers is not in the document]
dariuszseweryn: Indeed

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,45 @@
...
+
+@ConnectionScope
+class MtuWatcher extends Completable implements MtuProvider {
+
+    private final AtomicInteger currentMtuAtomicInteger;
+
+    @Inject
+    MtuWatcher(
+            final RxBleGattCallback rxBleGattCallback,
+            final AtomicInteger atomicInteger,
```
uKL: What is the reason of injecting an `AtomicInteger` rather than creating it in place? It seems to be only a container for a state. It's more like an implementation detail and should be hidden.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.inject.Named;
+import rx.Completable;
+import rx.CompletableSubscriber;
+import rx.Subscription;
+import rx.functions.Action1;
+
+@ConnectionScope
+class MtuWatcher extends Completable implements MtuProvider {
+
+    private final AtomicInteger currentMtuAtomicInteger;
```
uKL: Does it have to be atomic? Do we expect concurrent updates from different threads?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,45 @@
...
+    MtuWatcher(
+            final RxBleGattCallback rxBleGattCallback,
+            final AtomicInteger atomicInteger,
+            @Named(ConnectionComponent.NamedInts.GATT_MTU_MINIMUM) final int initialValue
+    ) {
+        super(new OnSubscribe() {
+            @Override
+            public void call(CompletableSubscriber completableSubscriber) {
+                Subscription mtuSubscription = rxBleGattCallback.getOnMtuChanged()
+                        .retry()
```
uKL: What is the scenario for `retry()` is required at all?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,45 @@
...
+                        .subscribe(new Action1<Integer>() {
+                            @Override
+                            public void call(Integer newMtu) {
+                                atomicInteger.set(newMtu);
+                            }
+                        });
+                completableSubscriber.onSubscribe(mtuSubscription);
```
uKL: Would it be updated if no one is listening to this completable?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
```diff
@@ -47,9 +47,11 @@ public RxBleConnection call() throws Exception {
                 final Observable<RxBleConnection> disconnectedErrorObservable = connectionComponent.gattCallback().observeDisconnect();
                 final DisconnectAction disconnect = connectionComponent.disconnectAction();
 
-                return newConnectionObservable
-                        .delaySubscription(connectedObservable)
-                        .mergeWith(disconnectedErrorObservable)
+                return Observable.merge(
+                        newConnectionObservable.delaySubscription(connectedObservable),
+                        disconnectedErrorObservable,
+                        connectionComponent.mtuWatcherCompletable().<RxBleConnection>toObservable()
```
uKL: This is acceptable but I'm thinking about some better way. This introduces API of a `Completable` that doesn't do anything but stays subscribed. Is there any good way to watch for connect/disconnect events and notify `MtuWatcher` to subscribe/unsubscribe from the MTU changes? This would be more descriptive that merging with a magic` Completable`.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.inject.Inject;
+import javax.inject.Named;
+import rx.Completable;
+import rx.CompletableSubscriber;
+import rx.Subscription;
+import rx.functions.Action1;
+
+@ConnectionScope
+class MtuWatcher extends Completable implements MtuProvider {
```
uKL: MtuWatcher or just MtuCache?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
dariuszseweryn: When the `BluetoothGattCallback.onMtuChanged()` would be called with `status != GATT_SUCCESS`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
dariuszseweryn: No. That is why it is subscribed in the `ConnectorImpl`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
dariuszseweryn: It is referenced in the `new OnSubscribe()` and it cannot be referenced there if it is a member of the `MtuWatcher`. The problem is that there is no easy way of extending a `Completable`. Maybe it is even an anti pattern as we can have independent `Completable` subscriptions and shared `AtomicInteger`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
dariuszseweryn: It has to be a reference so it could be referenced from the `OnSubscribe`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
dariuszseweryn: Currently it is watching the Mtu and providing the current value, name may be changed

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/ConnectorImpl.java
dariuszseweryn: I just try to avoid manual state management

uKL: Why is manual state management bad? I think it might be more descriptive.

If you want to stick with this solution I would name the method some other way. It should clearly say that it is responsible for the watcher state.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
uKL: ok

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/MtuWatcher.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.inject.Named;
+import rx.Observable;
+import rx.functions.Action1;
+import rx.subscriptions.SerialSubscription;
+
+@ConnectionScope
+class MtuWatcher implements ConnectionSubscriptionWatcher, MtuProvider, Action1<Integer> {
+
+    private Integer currentMtu;
+
```
uKL: Do we need empty lines here? In other places, we don't keep them.

PR #295, opened 2017-10-24
Added log whilst scan operation is requested to stop

uKL:


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/ScanOperation.java
```diff
@@ -33,6 +33,7 @@ final protected void protectedRun(final Emitter<SCAN_RESULT_TYPE> emitter, Queue
             emitter.setCancellation(new Cancellable() {
                 @Override
                 public void cancel() throws Exception {
+                    RxBleLog.v("Scan operation is requested to stop.");
```
dariuszseweryn: I am thinking wether it should be an `.i()` level. We could also use the same level log at the moment of starting the scan.

A `.d()` level log could be used in a specific scan operation (legacy, api18, api21) that was used

uKL: Good idea.

PR #298, opened 2017-10-24
Terminate ConnectionOperationQueue with proper exception

### LINKED ISSUE #297: Disconnection error is not properly routed
### Summary
When a connection error happens the `ConnectionOperationQueue` is always terminated with the `BleDisconnectedException` even when a different error has caused the disconnection
### Library version
`1.4.1`

<br />1. Connect to a device
<br />2. Make the `BluetoothGattCallback.onConnectionStateChange()` to be called with `status!=GATT_SUCCESS` so the operation would fail with `BleGattException(status, ON_CONNECTION_STATE)`
<br />3. Try to schedule a new operation and observe it will fail with `BleDisconnectedException` instead of the error from Step 2.

#### Actual result
`BleDisconnectedException`

#### Expected result
`BleGattException`

dariuszseweryn:
As long as the user is interested in the connection the `ConnectionOperationQueue` will check the `DisconnectionRouterOutput` for exceptions that cause disconnection. If this exception is emitted the `ConnectionOperationQueue` terminates itself with the emitted error. If the user unsubscribes from the `RxBleDevice.establishConnection()` then the `ConnectionOperationQueue` terminates itself with the `BleDisconnectedException` as it was before this change. Fixes #297

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectAction.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectAction.java
@@ -1,29 +1,20 @@
 package com.polidea.rxandroidble.internal.connection;
 
-import android.bluetooth.BluetoothDevice;
-import com.polidea.rxandroidble.exceptions.BleDisconnectedException;
 import com.polidea.rxandroidble.internal.operations.DisconnectOperation;
 import com.polidea.rxandroidble.internal.serialization.ClientOperationQueue;
-import com.polidea.rxandroidble.internal.serialization.ConnectionOperationQueue;
 import javax.inject.Inject;
-import rx.Subscription;
 import rx.functions.Actions;
 
 @ConnectionScope
-public class DisconnectAction implements ConnectionSubscriptionWatcher {
+class DisconnectAction implements ConnectionSubscriptionWatcher {
 
-    private final ConnectionOperationQueue connectionOperationQueue;
     private final ClientOperationQueue clientOperationQueue;
     private final DisconnectOperation operationDisconnect;
-    private final BluetoothDevice bluetoothDevice;
 
     @Inject
-    DisconnectAction(ConnectionOperationQueue connectionOperationQueue, ClientOperationQueue clientOperationQueue,
-                     DisconnectOperation operationDisconnect, BluetoothDevice bluetoothDevice) {
-        this.connectionOperationQueue = connectionOperationQueue;
+    DisconnectAction(ClientOperationQueue clientOperationQueue, DisconnectOperation operationDisconnect) {
         this.clientOperationQueue = clientOperationQueue;
         this.operationDisconnect = operationDisconnect;
-        this.bluetoothDevice = bluetoothDevice;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectAction.java
```diff
@@ -33,16 +24,12 @@ public void onConnectionSubscribed() {
...
-    private Subscription enqueueDisconnectOperation(DisconnectOperation operationDisconnect) {
-        return clientOperationQueue
+        clientOperationQueue
                 .queue(operationDisconnect)
                 .subscribe(
                         Actions.empty(),
                         Actions.<Throwable>toAction1(Actions.empty())
                 );
     }
+
```
uKL: Please remove empty line.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
```diff
@@ -17,26 +17,19 @@
...
-                disconnectionErrorRelay
-                        .flatMap(new Func1<BleException, Observable<?>>() {
-                            @Override
-                            public Observable<?> call(BleException e) {
-                                return Observable.error(e);
-                            }
-                        }),
+        disconnectionErrorOutputObservable = Observable.merge(
+                disconnectionErrorInputRelay,
                 adapterStateObservable
```
uKL: Can we have it extracted to `emitErrorWhenAdapterIsDisabled` or similar?

uKL: Shouldn't it rely on the initial state as well?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
```diff
@@ -63,31 +56,39 @@ public Boolean call(Boolean isAdapterUsable) {
...
     /**
-     * Method to be called whenever a BluetoothGattCallback.onConnectionStateChange() will get called with status != GATT_SUCCESS
-     *
-     * @param disconnectedGattException the exception that happened
+     * @inheritDoc
      */
-    void onGattConnectionStateException(BleGattException disconnectedGattException) {
-        disconnectionErrorRelay.call(disconnectedGattException);
+    @Override
+    public Observable<BleException> asExactObservable() {
```
uKL: What does it mean 'asExactObservable'? :) 😃  Could the method say something about the behavior?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/serialization/ConnectionOperationQueueImpl.java
```diff
@@ -16,25 +21,33 @@
...
 
-    private OperationPriorityFifoBlockingQueue queue = new OperationPriorityFifoBlockingQueue();
+    private final String deviceMacAddress;
+    private final DisconnectionRouterOutput disconnectionRouterOutput;
+    private Subscription disconnectionThrowableSubscription;
+    private final OperationPriorityFifoBlockingQueue queue = new OperationPriorityFifoBlockingQueue();
     private final Future<?> runnableFuture;
     private volatile boolean shouldRun = true;
-    private volatile BleDisconnectedException bleDisconnectedException = null;
+    private volatile BleException disconnectionException = null;
```
uKL: Effectively `volatile` is not required here because of synchronization. Having synchronized access guarantees that the other thread will read the value *after* the previous has written it.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/DisconnectionRouter.java
dariuszseweryn: Was not an issue till now. The DisconnectionRouter would not get created (@ConnectionScope) if the adapter is off

uKL: ok

Issue #279, opened 2017-09-08
Provide message for Exception classes


HolySamosa:
Some (or all?) of RxAndroidBle's Exception classes currently return `null` for `getMessage()`, making it difficult to provide meaningful error messages in the event of unexpected exceptions.

dariuszseweryn:
Hello @HolySamosa 
Could you give more context on how/when would you expect to use the value returned by `.getMessage()`?

dariuszseweryn:
@HolySamosa ping?

dariuszseweryn:
Closing due to inactivity. 

All exceptions are typed and can be checked in the sources what they mean. Showing contents of the `Exception.getMessage()` in the UI is a bad thing to do as most of the time it contains some description that is not understandable for the user. For the development purposes one can use `.toString()` which is implemented in all `BleException`s for quick debugging.

HolySamosa:
@dariuszseweryn Sorry for disappearing on this.  We were hit by Hurricane Irma just after this was filed, knocking out power and internet for the better part of a week.  I missed the earlier notifications and I'm just now catching up...

I ran into this issue w/ code for handling unexpected errors in my application.  While for the exception classes that most commonly occur the application provides a user-friendly message that doesn't make use of `.getMessage()`, in the event of unexpected errors there is an opportunity for the user to get more details about the error and in this case `getMessage()` is used. Also, this is a shared error handler at the presentation layer of my app, which knows nothing about **RxAndroidBle**.

Thanks for your patience and the great library. :-)

dariuszseweryn:
Sorry to hear that.

We will evaluate the idea of adding messages. As a temporary solution you could map the `RxAndroidBle` exceptions to your own ones at the boundary of the layer that knows about the library. :)

HolySamosa:
Mapping the exceptions is exactly what I am doing. :-)

And thanks for the kind words. I was fortunate in that we only were without power-- no damage to our home, just a mess of small tree limbs to clean up.

dariuszseweryn:
If you are mapping the exceptions then why the need for the messages? As far as I remember the library only uses exceptions that are declared in `/exceptions`.

dariuszseweryn:
Ping? :)

HolySamosa:
I'm mapping the "reason" values of `BleScanException`, as that's the case where my application was seeing unexpected errors.  I'm not mapping the others, as I haven't really ran into unexpected errors outside of scanning.

[...]

PR #301, opened 2017-10-30
Fix exceptions stack trace so it include message instead of "null"

connected-jallen:


dariuszseweryn:
Hello @connected-jallen!

Thank you for your pull request—it looks promising!
There are two things that make me think:
1. Now we loose the ability to call `(BleScanException| BleCannotSetCharacteristicNotificationException|BleDisconnectedException).toString()`. It would be a regression in terms of the funtionality
2. The exception message is usually a sentence that is easy to read by a human. The former implmenetation of `.toString()` gives all relevant information but does not make too much sense for a user that is not into BLE. Maybe the message could be just a sentence? i.e. having a `BleScanException` with a reason `LOCATION_SERVICES_DISABLED` could have a message like `Cannot start a scan due to Location Services being disabled`

What do you think? I look forward to discuss and merge your changes.

Josh-Allen-ConnectedLab:
@dariuszseweryn Look at the source to Throwable and you will see that toString() returns message. You can probably get rid of your classname in createMessage() because it's already included by Throwable toString. Did you not notice that whenever you logged your exceptions using Timber or Log that it always said "BleScanException: null"? The way you had it coded means we need a special case for all of your exceptions to include toString() as well as stacktrace.

dariuszseweryn:
That's true, we already have an opened issue for the messages #279 
I am thinking wether we should make the message more readable for non-tech savvy users. What do you think about it @connected-jallen ?
The class name indeed seems to be superfluous here.

connected-jallen:
@dariuszseweryn I made the toString() messages cleaner.

uKL:
It seems that this PR is a subset of #303, isn't it? I'll close this one in favor of #303. If you disagree please comment :) Thanks! Good job!

Issue #195, opened 2017-05-18
Characteristic change event not received 


gryffindor14:
### Summary
<br />
A characteristic change event is logged but never received through the subscribed observable when a CharacteristicChangeEvent is fired while the notification is being setup

#### Preconditions
<br />
A RxBleDevice is connected and has a characteristic with indication property
#### Steps to reproduce actual result<br />
1. connect to a BluetoothDevice<br />
2. setup indications<br />

```
Observable.combineLatest(
connection.setupIndication(UUID.fromString(00008a82-0000-1000-8000-00805f9b34fb)),
connection.setupIndication(UUID.fromString(00008a91-0000-1000-8000-00805f9b34fb)),
(indication1, indication2) -> { return indication1;})
.flatMap(observable-> observable)
.doOnNext(bytes -> { log.d(bytes) })
.subscribe();
```

3. As soon as onDescriptor write is fired a onChangeEvent happens that the observable never sees 

#### Actual result
<br />
 No log

#### Expected result
<br />
 Log of bytes

Actual Log:
Galaxy s6 6.0.1
```
D/RxBle#Radio: FINISHED RxBleRadioOperationConnect(149819350)
 D/BluetoothGatt: onClientConnParamsChanged() - Device=98:7B:F3:6B:F9:99 interval=6 status=0
D/RxBle#Radio:   QUEUED RxBleRadioOperationServicesDiscover(8272968)
 D/RxBle#Radio:  STARTED RxBleRadioOperationServicesDiscover(8272968)
 D/BluetoothGatt: discoverServices() - device: 98:7B:F3:6B:F9:99
 D/BluetoothGatt: onSearchComplete() = Device=98:7B:F3:6B:F9:99 Status=0
D/RxBle#BluetoothGatt: onServicesDiscovered status=0
D/BluetoothGatt: setCharacteristicNotification() - uuid: 00008a82-0000-1000-8000-00805f9b34fb enable: true
 D/RxBle#Radio:   QUEUED RxBleRadioOperationDescriptorWrite(72475096)
D/RxBle#Radio: FINISHED RxBleRadioOperationServicesDiscover(8272968)
D/RxBle#Radio:  STARTED RxBleRadioOperationDescriptorWrite(72475096)
D/BluetoothGatt: onClientConnParamsChanged() - Device=98:7B:F3:6B:F9:99 interval=39 status=0
[...]

PR #335, opened 2017-12-05
Start observing characteristic change before setup BLE notfication

mswokk:
Change sequence of registering observer and request notification to BLE device.
Fixes issue that missing first characteristic change by Notification request.

See issue #195


mswokk:
Close and will create a new request with error handling

PR #336, opened 2017-12-06
Register notification callback before observe it

mswokk:
Change sequence of registering observer and setting up the notification to BLE device.
Fixes issue that missing first characteristic change by Notification request.

See issue #195

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/NotificationAndIndicationManager.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble/internal/connection/NotificationAndIndicationManager.java
@@ -1,10 +1,10 @@
 package com.polidea.rxandroidble.internal.connection;
 
-
 import android.bluetooth.BluetoothGatt;
 import android.bluetooth.BluetoothGattCharacteristic;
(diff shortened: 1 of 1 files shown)


dariuszseweryn:
There are few things to consider:

* The user loses the ability to tell when is the exact moment the notification is successfully setup. There are cases where users need to wait for the notification to setup before proceeding with next BLE operations.
* This exact behaviour is already possible to obtain by using `NotificationSetupMode.COMPAT` while setting the notification/indication. It makes the function to not write the Client Characteristic Config descriptor. Then the user just need to write the descriptor manually after they start observing notifications
* There is an already merged PR Polidea/react-native-ble-plx#195 which enables this exact behaviour without changes in RxAndroidBle

dariuszseweryn:
As the issue is already fixed in `react-native-ble-plx` and there are doubts if the proposed solution should be integrated with the current API of the `RxAndroidBle`.

The proposed behaviour will probably get implemented with [the new API](https://github.com/Polidea/RxAndroidBle/pull/239) as an alternative for the current one but it will need some work.

Thank you for your effort and have a happy new year using the libraries! ;)

PR #354, opened 2018-01-10
Added dagger shadows to gitignore.

dariuszseweryn:
I think it may have sense. Feel free to comment if it should not be ignored.


uKL:
It may lead to some confusion if for some reason we'd need to add something to the modules. Maybe let's ignore `build/*` withing those?

PR #357, opened 2018-01-18
Added retry mechanism parameter for LongWriteOperation

martiwi:
PR for issue https://github.com/Polidea/RxAndroidBle/issues/352


rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -62,12 +67,14 @@
         this.bluetoothGattCharacteristic = bluetoothGattCharacteristic;
         this.batchSizeProvider = batchSizeProvider;
         this.writeOperationAckStrategy = writeOperationAckStrategy;
+        this.writeOperationRetryStrategy = writeOperationRetryStrategy;
         this.bytesToWrite = bytesToWrite;
+        this.retryCounter = 0;
```
dariuszseweryn: I am not a big fan of variables that are changed as a side effect. This usually makes room for a mistake and makes the code harder to trace.

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -442,6 +450,9 @@ public String toString() {
      */
     Observable<byte[]> writeDescriptor(@NonNull BluetoothGattDescriptor descriptor, @NonNull byte[] data);
 
+    interface WriteOperationRetryStrategy extends Func2<Integer, Throwable, Boolean> {
```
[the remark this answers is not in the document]
martiwi: Ok.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: You mean we should not have a separated counter in this Operation class at all?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -87,9 +94,11 @@ protected void protectedRun(final Emitter<byte[]> emitter, final QueueReleaseInt
                         timeoutObservable,
                         timeoutConfiguration.timeoutScheduler
                 )
+                .doOnNext(resetRetryCounter())
                 .repeatWhen(bufferIsNotEmptyAndOperationHasBeenAcknowledgedAndNotUnsubscribed(
                         writeOperationAckStrategy, byteBuffer, emitterWrapper
                 ))
+                .retry(retryStrategy(byteBuffer, batchSize))
```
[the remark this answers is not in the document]
martiwi: Indeed, that makes sense.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
uKL: I believe that Darek meant putting a reset in a `doOnNext`. AFAIK you can safely reset the counter before the rx subscription (around line 86.)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +218,30 @@ public Boolean call(Object emission) {
...
+        };
+    }
+
+    private Func2<Integer, Throwable, Boolean> retryStrategy(final ByteBuffer byteBuffer, final int batchSize) {
+        return new Func2<Integer, Throwable, Boolean>() {
+            @Override
+            public Boolean call(Integer integer, Throwable throwable) {
+                // Individual counter for each batch payload.
+                retryCounter++;
+                final Boolean retry = writeOperationRetryStrategy.call(retryCounter, throwable);
```
uKL: The biggest question here is to:
1) Define what data one will need to decide on a retry
2) Smart way to extend that data without breaking the API in the future (DTO? @dariuszseweryn)

dariuszseweryn: Yup, probably we should go for `WriteOperationRetryStrategy` with something like:
```
class LongWriteFailure {
  final int batchNumber;
  // final int retryCount; –> this should not be needed as it is trivial for the user to manage
  final BleException cause;
}
```

uKL: You're reading in my mind :)

martiwi: I came up with the following `retryWhen()` function:
```java
private static Func1<Observable<? extends Throwable>, Observable<?>> retryOperationStrategy(
            final WriteOperationRetryStrategy writeOperationRetryStrategy,
            final ByteBuffer byteBuffer,
            final int batchSize) {
        return new Func1<Observable<? extends Throwable>, Observable<?>>() {
            @Override
            public Observable<?> call(Observable<? extends Throwable> observable) {
                return observable
                        .flatMap(applyRetryStrategy())
                        .map(replaceByteBufferPositionForRetry());
            }

            @NonNull
            private Func1<Throwable, Observable<WriteOperationRetryStrategy.LongWriteFailure>> applyRetryStrategy() {
                return new Func1<Throwable, Observable<WriteOperationRetryStrategy.LongWriteFailure>>() {
                    @Override
                    public Observable<WriteOperationRetryStrategy.LongWriteFailure> call(Throwable cause) {
                        if (!(cause instanceof BleException)) {
                            return Observable.error(cause);
                        }

                        final int failedBatchNumber = (byteBuffer.position() - batchSize) / batchSize;
                        final WriteOperationRetryStrategy.LongWriteFailure longWriteFailure =
                                new WriteOperationRetryStrategy.LongWriteFailure(failedBatchNumber, (BleException) cause);

                        return writeOperationRetryStrategy.call(Observable.just(longWriteFailure));
                    }
                };
            }

            @NonNull
            private Func1<WriteOperationRetryStrategy.LongWriteFailure, Object> replaceByteBufferPositionForRetry() {
                return new Func1<WriteOperationRetryStrategy.LongWriteFailure, Object>() {
                    @Override
                    public Object call(WriteOperationRetryStrategy.LongWriteFailure longWriteFailure) {
                        final int newBufferPosition = longWriteFailure.getBatchNumber() * batchSize;
                        byteBuffer.position(newBufferPosition);
                        return longWriteFailure;
                    }
                };
            }
        };
    }
```

dariuszseweryn: If you could push another commit to this PR I could then comment on specific lines

martiwi: Sure, done!

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +213,48 @@ public Boolean call(Object emission) {
...
+                    public Observable<WriteOperationRetryStrategy.LongWriteFailure> call(Throwable cause) {
+                        if (!(cause instanceof BleException)) {
+                            return Observable.error(cause);
+                        }
+
+                        final int failedBatchNumber = (byteBuffer.position() - batchSize) / batchSize;
+                        final WriteOperationRetryStrategy.LongWriteFailure longWriteFailure =
+                                new WriteOperationRetryStrategy.LongWriteFailure(failedBatchNumber, (BleException) cause);
+
+                        return writeOperationRetryStrategy.call(Observable.just(longWriteFailure));
```
dariuszseweryn: Not an ideal solution since the `writeOperationRetryStrategy` will get called each time the a new failure will happen where is could be called only once per subscription. Checkout how the `Observable.compose()` function or `bufferIsNotEmptyAndOperationHasBeenAcknowledgedAndNotUnsubscribed` works.

Basically this: 
```
final int failedBatchNumber = (byteBuffer.position() - batchSize) / batchSize;
                         final WriteOperationRetryStrategy.LongWriteFailure longWriteFailure =
                                 new WriteOperationRetryStrategy.LongWriteFailure(failedBatchNumber, (BleException) cause);
```
could be separated to a `Observable.map()` and the `writeOperationRetryStrategy` could be treated as a `Observable.Transformer` (see `.compose()`)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +213,48 @@ public Boolean call(Object emission) {
...
+                    }
+                };
+            }
+
+            @NonNull
+            private Func1<WriteOperationRetryStrategy.LongWriteFailure, Object> replaceByteBufferPositionForRetry() {
```
dariuszseweryn: This could be a simple `.onNext()` as it only passes the `longWriteFailure`

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
```diff
@@ -160,6 +180,45 @@ public String toString() {
         Observable<byte[]> build();
     }
 
+    interface WriteOperationRetryStrategy extends Func1<Observable<WriteOperationRetryStrategy.LongWriteFailure>,
+            Observable<WriteOperationRetryStrategy.LongWriteFailure>> {
+
+        class LongWriteFailure {
+
+            final int batchNumber;
+            final BleException cause;
```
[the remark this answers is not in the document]
martiwi: Ok for `BleGattCharacteristicException` but they are already both final :)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: Ok.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: How close am I with something like that :) :
```java
            @Override
            public Observable<?> call(Observable<? extends Throwable> emittedOnWriteFailure) {
                return writeOperationRetryStrategy.call(
                        emittedOnWriteFailure.map(mapToLongWriteFailureObject())
                )
                        .doOnNext(replaceByteBufferPositionForRetry());
            }
```
No other `Throwable` than `BleException` can be emitted?

rxandroidble/src/main/java/com/polidea/rxandroidble/RxBleConnection.java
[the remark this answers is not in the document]
dariuszseweryn: Yup. I must be tired...

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
dariuszseweryn: No other than `BleGattCharacteristicException` can be retried... Again, new commit in this PR makes it easier to see the whole picture. :)

martiwi: Sorry, I've pushed :)

dariuszseweryn: Yup, I have seen—just have a bit of work now. Will get back to you as soon as possible. 👍

dariuszseweryn: I have checked out your branch and I hope I will have a moment Tomorrow to play with it and see if I can make it a little bit nicer. I cannot think of how to explain this so I will try to show it ;)

dariuszseweryn:
I didn't find a moment Today to play with it. Tomorrow should be the day

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,58 @@ public Boolean call(Object emission) {
...
+                        .compose(writeOperationRetryStrategy)
+                        .doOnNext(replaceByteBufferPositionForRetry());
+            }
+
+            @NonNull
+            private Func1<Throwable, WriteOperationRetryStrategy.LongWriteFailure> toLongWriteFailureObject() {
+                return new Func1<Throwable, WriteOperationRetryStrategy.LongWriteFailure>() {
+                    @Override
+                    public WriteOperationRetryStrategy.LongWriteFailure call(Throwable throwable) {
+                        final int failedBatchNumber = (byteBuffer.position() - batchSize) / batchSize;
```
dariuszseweryn: This calculation will be wrong if the last batch will not be equal size to default `batchSize`. Ideally we should be passing the batch size to this function from the writeBatchAndObserve().

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,58 @@ public Boolean call(Object emission) {
...
+                        return new WriteOperationRetryStrategy.LongWriteFailure(
+                                failedBatchNumber,
+                                (BleGattCharacteristicException) throwable
+                        );
+                    }
+                };
+            }
+
+            @NonNull
+            private Action1<WriteOperationRetryStrategy.LongWriteFailure> replaceByteBufferPositionForRetry() {
```
dariuszseweryn: Maybe `repositionByteBufferForRetry()`?

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,58 @@ public Boolean call(Object emission) {
...
+                    @Override
+                    public void call(WriteOperationRetryStrategy.LongWriteFailure longWriteFailure) {
+                        final int newBufferPosition = longWriteFailure.getBatchNumber() * batchSize;
+                        byteBuffer.position(newBufferPosition);
+                    }
+                };
+            }
+
+            @NonNull
+            private Func1<Throwable, Observable<Throwable>> canRetryError() {
```
dariuszseweryn: Maybe `toLongWriteFailureOrError()`? It would then look like `.flatMap(toLongWriteFailureOrError())`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,58 @@ public Boolean call(Object emission) {
...
+            final WriteOperationRetryStrategy writeOperationRetryStrategy,
+            final ByteBuffer byteBuffer,
+            final int batchSize) {
+        return new Func1<Observable<? extends Throwable>, Observable<?>>() {
+            @Override
+            public Observable<?> call(Observable<? extends Throwable> emittedOnWriteFailure) {
+                return emittedOnWriteFailure
+                        .flatMap(canRetryError())
+                        .map(toLongWriteFailureObject())
+                        .compose(writeOperationRetryStrategy)
```
dariuszseweryn: This is what I have changed which makes it easier to read in my opinion. I will refactor the ACK strategy one day...

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,58 @@ public Boolean call(Object emission) {
             }
         };
     }
+
+    private static Func1<Observable<? extends Throwable>, Observable<?>> retryOperationStrategy(
```
dariuszseweryn: Maybe `errorIsRetriableAndAccordingTo(...)`? Then we could have `.retryWhen(errorIsRetriableAndAccordingTo(writeOperationRetryStrategy,...))`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: You mean merge those two?
```
.flatMap(canRetryError())
.map(toLongWriteFailureObject())
```

dariuszseweryn: Yes, as the the `.map()` contains a pretty straightforward logic. Maybe `toLongWriteFailureIfRetryable()` (not sure about the spelling)

martiwi: I've split them based on comment: https://github.com/Polidea/RxAndroidBle/pull/357#discussion_r162981644 :)
Haha actually, I was wondering about the same thing, maybe `toLongWriteFailureIfCanBeRetried()`?

dariuszseweryn: It was exactly more about using `.compose(writeOperationRetryStrategy)`

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: So `writeBatchAndObserve()` should emit a DTO instead of a `ByteAssociation<UUID>` like:
```
class WriteBatch {
    final ByteAssociation<UUID> byteAssociation;
    final int batchSize;
}
```

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: Ok :)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: Ok.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: Ok.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
martiwi: Indeed, but what advantage compared to something like that instead:
```
if (byteBuffer.hasRemaining()) {
    return (byteBuffer.position() - batchSize) / batchSize;
} else {
    return byteBuffer.position() / batchSize;
}
```

dariuszseweryn: Ideally we should emit the batches to write `byte[]` which should be indexed. Those pairs of bytes and indexes should then be written (and retried) accordingly to the strategy to be in line with in as immutable approach as possible. I am not sure how would the performance turn up...

martiwi:
I have pushed my latest modifications accordingly to your remarks.

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,56 @@ public Boolean call(Object emission) {
...
+                    public void call(WriteOperationRetryStrategy.LongWriteFailure longWriteFailure) {
+                        final int newBufferPosition = longWriteFailure.getBatchNumber() * batchSize;
+                        byteBuffer.position(newBufferPosition);
+                    }
+                };
+            }
+
+            private int calculateFailedBatchNumber(ByteBuffer byteBuffer, int batchSize) {
+                if (byteBuffer.hasRemaining()) {
+                    return (byteBuffer.position() - batchSize) / batchSize;
```
dariuszseweryn: `(byteBuffer.position() / batchSize) - 1` seems slightly easier to understand for me :)

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +214,56 @@ public Boolean call(Object emission) {
...
+            final WriteOperationRetryStrategy writeOperationRetryStrategy,
+            final ByteBuffer byteBuffer,
+            final int batchSize) {
+        return new Func1<Observable<? extends Throwable>, Observable<?>>() {
+            @Override
+            public Observable<?> call(Observable<? extends Throwable> emittedOnWriteFailure) {
+                return emittedOnWriteFailure
+                        .flatMap(toLongWriteFailureOrError())
+                        .compose(writeOperationRetryStrategy)
+                        .doOnNext(repositionByteBufferForRetry());
```
dariuszseweryn: This line should go to above the `.compose()` as the user could potentially emit a different `LongWriteFailure` than they were passed which could flaw the reposition of the buffer.

dariuszseweryn:
Hello @martiwi - any news? Maybe I can help you with something?
I see this feature just lacking some unit tests and it would be production ready for `1.5.0` release

martiwi:
Hi there,
I didn't had enough time available to understand correctly how your unit test are designed and how they are working. I've been able to update the current unit tests to have latest changes of this feature. But to test that retry strategy is properly working, I did not managed to. Any help is welcome to speed up on this. Thanks.

uKL:
Hey @martiwi !

Could you tick "Allow edits from maintainers" in your PR? I not able to contribute to unit tests.

martiwi:
Hey @uKL it was already selected, @dariuszseweryn was able to publish before. I've unselected it, and re-selected it. Is it better now ?

uKL:
Ok, I was doing it wrong. :)

martiwi:
Hi @dariuszseweryn - I've added one unit test but I am not sure I am using it right as when running the test in debug mode, the second batch is failing to write but retryWhen Observable is triggered twice. Meaning `Characteristic#setValue()` is called with first batch number and not the second. Can you check what am I doing wrong? Thanks.

dariuszseweryn:
Hello @martiwi 
You can check if the whole chain has not failed at this point by using `testSubscriber.assertNoErrors()`. Additionally in the current test all batches after the first failed one will not be successfully completed as the subject is called with `.onError()`

uKL:
Great job @martiwi! 

I've added some tests and fixed two issues + a documentation update. 

- Added tests for retries during long write operations
- Fixed issue that caused the last batch to be written with incorrect data (empty byte array)
- Added support for errors that were occurring when the operation was started (not from the gatt callback)
- Fixed documentation incorrectly stating that the particular batch is skipped

rxandroidble/src/test/groovy/com/polidea/rxandroidble/internal/operations/OperationCharacteristicLongWriteTest.groovy
```diff
@@ -203,50 +203,164 @@ public class OperationCharacteristicLongWriteTest extends Specification {
...
+
+        testSubscriber.assertValueEquals([0x1, 0x1, 0x2, 0x2, 0x3, 0x3] as byte[])
+        testSubscriber.assertCompleted()
+    }
+
+    def "attempt to rewrite the failed batch if the strategy has emitted the LongWriteFailure - last batch"() {
+        given:
+        this.writeOperationRetryStrategy = givenWillRetryWriteOperation()
+        this.writeOperationAckStrategy = new ImmediateSerializedBatchAckStrategy()
+        prepareObjectUnderTest(2, [0x1, 0x1, 0x2, 0x2, 0x3, 0x3] as byte[])
```
dariuszseweryn: Could you make the last batch to have an uneven number of bytes? I think we can have a wrong `ByteBuffer` positioning on the last batch if is not full

rxandroidble/src/main/java/com/polidea/rxandroidble/internal/operations/CharacteristicLongWriteOperation.java
```diff
@@ -209,4 +215,52 @@ public Boolean call(Object emission) {
...
+                    @Override
+                    public void call(WriteOperationRetryStrategy.LongWriteFailure longWriteFailure) {
+                        final int newBufferPosition = longWriteFailure.getBatchNumber() * batchSize - batchSize;
+                        byteBuffer.position(newBufferPosition);
+                    }
+                };
+            }
+
+            private int calculateFailedBatchNumber(ByteBuffer byteBuffer, int batchSize) {
+                return (int) Math.ceil(byteBuffer.position() / (float) batchSize);
```
dariuszseweryn: doesn't it return a failed batch number + 1?

dariuszseweryn: Having a [1,2,3,4,5] and 2 bytes per batch
If the first batch fails the `byteBuffer.position()` would return 2
2/2 == 1 and the failed batch index should be 0

uKL: Data size 5, batch size 2, failure at the last batch (3rd)
(int) Math.ceil(5 / (float) 2) => 3

Data size 6, batch size 2, failure at the last batch (3rd)
(int) Math.ceil(5 / (float) 2) => 3

Data size 5, batch size 3, failure at the last batch (2nd)
(int) Math.ceil(5 / (float) 2) => 2

Is this what you wanted to check?

dariuszseweryn: `calculateFailedBatchNumber` is misleading in this situation, either we should return
`(int) Math.ceil(byteBuffer.position() / (float) batchSize) - 1`
or name the function differently?

uKL: It returns the "number" which I assumed starts from number one. `LongWriteFailure` also assumes "number". Do you think we should switch to zero-based indexes?

dariuszseweryn: Yup. This is user (developer) facing data
I think we should to zero-based indexes and rename the function to `calculateFailedBatchIndex` and `longWriteFailure.getBatchIndex()` respectively

uKL:
Dear Contributor,

PR #361, opened 2018-02-07
`autoConnect`—a better description

dariuszseweryn:
Added better description of how `autoConnect` flag works within this library.


RobLewis:
Sorry, I don't know how to edit the files directly, but I have a couple of style comments: 

In ReadMe: 
I would change this: 
Unlike the native Android API if `autoConnect=true` while using this library there will be NO attempts to automatically reconnect if the original connection will get lost.
to this: 
Unlike the native Android API, if `autoConnect=true` while using this library there will be NO attempts to automatically reconnect if the original connection is lost. 

In RxBleDevice.java: 
I would substitute this: 
If false, the connection will fail with {@link com.polidea.rxandroidble.exceptions.BleGattException} if
+     *                    the device is not in range after a 30-second timeout. If true, the connection will be pending indefinitely.
+     *                    Unlike the native Android API, if set to true and the connection is lost there will NOT be any attempt to
+     *                    reconnect unless explicitly resubscribed.

dariuszseweryn:
Splendid! Thank you for the suggestions @RobLewis I have incorporated all of them 👍

PR #384, opened 2018-03-08
Rxjava2.0 reconnect crash issue #383

eimermusic:
Project sample app modified to induce the issue described in https://github.com/Polidea/RxAndroidBle/issues/383 .

@dariuszseweryn I hope you can see a clear-cut problem with reconnecting this way... I can imagine that some of the disposal work might not be done?


eimermusic:
Seems like you can also get this exception by simply attempting to connect and re-connect in this way to a MAC address that you cannot reach. No need to actually successfully connect first.

uKL:
I'm going to close this PR as this is a reproducibility sample. I'll follow up with an explanation in the issue.

PR #385, opened 2018-03-08
Fix typo in gradle dependency

timmiller17:


timmiller17:
I'm not sure that I made the right correction, but I wanted to at least highlight that it seemed broken the way it was.

uKL:
Thanks for pointing that out. I was aware anyway. There are more places requiring an update, I'll update them once this is merged to master.

timmiller17:
Sounds good.  Hopefully it gets fixed soon.  It took me too much time to figure out that was the source of my problems trying to get the rxjava2.0 branch working.  :)

PR #389, opened 2018-03-12
[#383] Due to a way how RxJava2 handles undeliverable exceptions smal…

uKL:
…l changes to emitter based observers were required.


NitroG42:
I might be wrong but it seems to me that this merged commit is not in the 1.5.0 release, I am wrong ?
Asking because I have the aforementionned issue.

dariuszseweryn:
It is, just there were some other minor changes in the code. You can see it in the commit list in [the RxJava core PR](https://github.com/Polidea/RxAndroidBle/pull/362/commits)

PR #393, opened 2018-03-15
[#383] Trying to fix undeliverable exception when disposing MTU watcher.

uKL:


rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/connection/MtuWatcher.java
```diff
@@ -1,26 +1,33 @@
...
     MtuWatcher(
             final RxBleGattCallback rxBleGattCallback,
             @Named(ConnectionComponent.NamedInts.GATT_MTU_MINIMUM) final int initialValue
     ) {
-        this.mtuObservable = rxBleGattCallback.getOnMtuChanged().retry();
+        this.mtuObservable = rxBleGattCallback.getOnMtuChanged()
+                .retry(new Predicate<Throwable>() {
+                    @Override
+                    public boolean test(Throwable throwable) throws Exception {
+                        return !(throwable instanceof BleDisconnectedException);
```
dariuszseweryn: Actually it could only be the `BleGattException` with `BleGattOperationType==ON_MTU_CHANGED`

Issue #407, opened 2018-04-01
Scan results list looks scrambled due to RecyclerView animations when many devices present


kaiaai:
### Summary
List of devices in sample app shows scrambled when there are too many BLE devices.
This is caused by RecyclerView animations, especially when user scrolls through long list of scanned devices.

This is a purely esthetical issue, nothing wrong with functionality.
To fix, add recyclerView.setItemAnimator(null)
````
    private void configureResultList() {
        recyclerView.setHasFixedSize(true);
        recyclerView.setItemAnimator(null);
````

### Library version
Release 1.5.0

#### Preconditions
1) Many BLE devices present during scan, enough to fill more than entire screen in sample scan activity.
2) user scrolls list up/down

#### Steps to reproduce actual result
<br />1. Find a place having many BLE around (Starbucks works)
<br />2. Run sample app scan
<br />3. Scroll through list of scanned devices
#### Minimum code snippet reproducing the issue
Release 1.5.0

#### Logs from the application running with setting `RxBleLog.setLogLevel(RxBleLog.VERBOSE)`
N/A

#### Actual result
![animations_enabled](https://user-images.githubusercontent.com/33589365/38177279-b7999b72-35b3-11e8-9e7f-d59dcb0aa7fe.png)

#### Expected result
![animations_disabled](https://user-images.githubusercontent.com/33589365/38177308-1f61e426-35b4-11e8-9a4a-5d8a4a803dbf.png)

dariuszseweryn:
Hello,

Thanks for pointing that out!

Unfortunately we currently have limited time for the development. If you could create a PR we would be happy to merge it in right away!

Best Regards

PR #409, opened 2018-04-06
Removed animation from sample scanning result list

kaiaai:
Fix for issue #407 . Disable RecyclerView animations messing up screen when user scrolls through long device list during scan.


uKL:
Thanks!

Issue #412, opened 2018-04-10
Add Priority on CustomOperation


GuillaumeBo:
I am using CustomOperation to refresh Gatt in case of write/read fail and retry right after.
The issue is that CustomOperation is added to the queue with Normal priority, so if i launch 3 read operation for exemple. The first one fails -> i want to refresh Gatt, the custom operation will be executed after the 2 others read operations (that will fail too, same issue as the first).

The expected behaviour would be to assign High priority to this custom operation, in this case the refresh will be executed right after the first read operation and will benefit to the two others.

maybe a overridable method as `definedPriority()` of QueueOperation class ?

PR #414, opened 2018-04-10
Add possibility to modify custom operation priority #412

GuillaumeBo:
reference to #412


GuillaumeBo:
Thanks you, it's ok for CLA, i already signed it, but i commited with wrong email, now it's fixed :)

PR #417, opened 2018-04-16
Added possibility to change default connection timeout.

ihenchi:


rxandroidble/src/main/java/com/polidea/rxandroidble2/ConnectionSetup.java
```diff
@@ -32,19 +33,26 @@
...
+     */
+    public final Timeout connectTimeout;
 
-    private ConnectionSetup(boolean autoConnect, boolean suppressOperationCheck, Timeout operationTimeout) {
+    private ConnectionSetup(boolean autoConnect, boolean suppressOperationCheck, Timeout operationTimeout,
+                            Timeout connectTimeout) {
         this.autoConnect = autoConnect;
         this.suppressOperationCheck = suppressOperationCheck;
         this.operationTimeout = operationTimeout;
+        this.connectTimeout = connectTimeout;
```
piotrek1543: It's just a detail, but I wonder if instead of `connectTimeout` and `DEFAULT_CONNECT_TIMEOUT` shouldn't be `connectionTimeout` and `DEFAULT_CONNECTION_TIMEOUT` like there already are `DEFAULT_OPERATION_TIMEOUT` and `operationTimeout`

rxandroidble/src/main/java/com/polidea/rxandroidble2/ConnectionSetup.java
```diff
@@ -87,8 +95,19 @@ public Builder setOperationTimeout(Timeout operationTimeout) {
             return this;
         }
 
+        /**
+         * @param connectTimeout Timeout after which the connection will be considered as broken. Eventually the connection
+         *                         will be canceled and removed from queue. Keep in mind that it will cancel the library's connection
+         *                         only and may leave Android's BLE stack in an inconsistent state.
+         * @return this builder instance
+         */
+        public Builder setConnectTimeout(Timeout connectTimeout) {
```
piotrek1543: what about `setConnectionTimeout` (as you have before `setOperationTimeout()`)?

rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/RxBleDeviceImpl.java
```diff
@@ -67,6 +67,19 @@
         return establishConnection(options);
     }
 
+    @Override
+    public Observable<RxBleConnection> establishConnection(final boolean autoConnect,
+                                                           final Timeout operationTimeout,
+                                                           final Timeout connectTimeout) {
+        ConnectionSetup options = new ConnectionSetup.Builder()
```
piotrek1543: Nice.

rxandroidble/src/main/java/com/polidea/rxandroidble2/RxBleDevice.java
```diff
@@ -109,6 +109,49 @@
      */
     Observable<RxBleConnection> establishConnection(boolean autoConnect, @NonNull Timeout operationTimeout);
 
+    /**
```
uKL: Could you also extend the `OperationConnectTest` so it tests configurable timeouts? Currently it only has the `should emit BluetoothGattCallbackTimeoutException with a valid mac address on CallbackTimeout`

rxandroidble/src/main/java/com/polidea/rxandroidble2/RxBleDevice.java
```diff
@@ -109,6 +109,49 @@
...
+     *                         {@link android.bluetooth.BluetoothDevice#connectGatt(Context, boolean, BluetoothGattCallback)}
+     *                         autoConnect flag.
+     *                         If false, the connection will fail with {@link com.polidea.rxandroidble2.exceptions.BleGattException} if
+     *                         the device is not in range after a 30-second timeout. If true, the connection will be pending indefinitely.
+     *                         Unlike the native Android API, if set to true and the connection is lost there will NOT be any attempt to
+     *                         reconnect unless explicitly resubscribed.
+     * @param operationTimeout Timeout configuration after which the operation will be considered as broken. Eventually the operation
+     *                         will be canceled and removed from queue. Keep in mind that it will cancel the library's operation
+     *                         only and may leave Android's BLE stack in an inconsistent state.
+     * @param connectTimeout   Timeout configuration after which the connection will be considered as broken. Eventually the connection
```
uKL: > Keep in mind that it will cancel the library's connection only and may leave Android's BLE stack in an inconsistent state.

This does not apply to connections. Each connection attempt is disconnected and then closed when the connection subscription is released. Please update the documentation.

rxandroidble/src/test/groovy/com/polidea/rxandroidble2/internal/RxBleDeviceProviderTest.groovy
```diff
@@ -39,6 +39,13 @@ class RxBleDeviceProviderTest extends Specification {
         io.reactivex.Observable<RxBleConnection> establishConnection(boolean autoConnect, Timeout operationTimeoutSetup) {
             establishConnection(autoConnect)
         }
+
+//        @Override
```
uKL: I suppose that the `@Override` shouldn't be commented out. It applies to the method below which is related to the new API which is not yet public.

rxandroidble/src/main/java/com/polidea/rxandroidble2/ConnectionSetup.java
dariuszseweryn: Maybe `setConnectingTimeout`? `Connection` may be understood as a whole but `Connecting` is the process of establishing a `Connection`?

piotrek1543: @dariuszseweryn you're right. `setConnectingTimeout` would be better as it is more accurate.

rxandroidble/src/main/java/com/polidea/rxandroidble2/RxBleDevice.java
ihenchi: @uKL the configurable connecting timeout is set at the `RxBleDevice` level. I am not sure how I can extend `OperationConnectTest`.

rxandroidble/src/main/java/com/polidea/rxandroidble2/RxBleDevice.java
```diff
@@ -109,6 +109,48 @@
...
+     *                         autoConnect flag.
+     *                         If false, the connection will fail with {@link com.polidea.rxandroidble2.exceptions.BleGattException} if
+     *                         the device is not in range after a 30-second timeout. If true, the connection will be pending indefinitely.
+     *                         Unlike the native Android API, if set to true and the connection is lost there will NOT be any attempt to
+     *                         reconnect unless explicitly resubscribed.
+     * @param operationTimeout Timeout configuration after which the operation will be considered as broken. Eventually the operation
+     *                         will be canceled and removed from queue. Keep in mind that it will cancel the library's operation
+     *                         only and may leave Android's BLE stack in an inconsistent state.
+     * @param connectingTimeout   Timeout configuration after which the connection will be considered as broken. Eventually the connection
+     *                         attempt will be disconnected and then closed when the connection subscription is released.
```
dariuszseweryn: I would add here information that it is only taken onto account when used with `autoConnect=false`.

rxandroidble/src/main/java/com/polidea/rxandroidble2/RxBleDevice.java
dariuszseweryn: It is set on `RxBleDevice` level but ultimately it ends in `ConnectOperation` and should be tested properly in `def "should emit BluetoothGattCallbackTimeoutException with a valid mac address on CallbackTimeout"` test in `OperationConnectTest:144` where there is a hardcoded value `35`. It should be verified that the timeout will happen exactly as specified in `OperationTimeoutConfiguration`

PR #427, opened 2018-05-18
Fix build broken by #414 by adding missing implementation

bleeding182:
Mock module wouldn't build since the interface changed with #414 is not implemented there


dariuszseweryn:
Thanks! Going for the other PRs (not sure if I'll manage all Today)

PR #428, opened 2018-05-18
Fix deadlock reported by #426 when FIFORunnable is disposed before running

### LINKED ISSUE #426: ConnectOperation deadlocks ClientOperationQueueImpl
### Summary

I am testing with multiple BT devices, opening & closing connections, etc. Every once in a while everything stops working.

When attaching a Debugger and pausing the program I can see that "Thread-2" (or whatever its current name) is waiting for the release of `QueueSemaphore`. It is effectively deadlocked. The operation it is waiting for is `ConnectOperation`.

The semaphore does not properly get released when the observer was disposed. It also seems a similar deadlock could occur when an exception is raised.

### Library version

    implementation 'com.polidea.rxandroidble2:rxandroidble:1.5.0'

#### Preconditions

Open/Close multiple connections in quick succession

It occurs randomly when connecting/disconnecting etc, as such I can't provide a code sample to reproduce the issue. It happens regularly on a Nexus 6P which seems generally rather slow opening a new connection.

### Error Description

The Deadlock occurs when the observer was disposed before the operation is run.

At `FIFORunnableEntry.java:74` the else branch will _not_ release the semaphore, thus deadlocking the queue.

It also seems that `QueueOperation.java:43` and `QueueOperation.java:45` should release the semaphore, to prevent a deadlock here after ignoring any potential errors.

---

It seems that this issue can be fixed by adding a call to `semaphore.release()` in the 3 mentioned locations.

bleeding182:
As reported we should release the semaphore when we don't execute the operation.
Fixes #426

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/serialization/FIFORunnableEntry.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/serialization/FIFORunnableEntry.java
@@ -35,45 +35,46 @@ class FIFORunnableEntry<T> implements Comparable<FIFORunnableEntry> {
     }
 
     public void run(QueueSemaphore semaphore, Scheduler subscribeScheduler) {
+
+        if (operationResultObserver.isDisposed()) {
+            RxBleLog.d("Operation was about to be run but the observer was already disposed: " + operation);
+            semaphore.release();
+            return;
+        }
+
         /*
          * In some implementations (i.e. Samsung Android 4.3) calling BluetoothDevice.connectGatt()
          * from thread other than main thread ends in connecting with status 133. It's safer to make bluetooth calls
          * on the main thread.
          */
 
-        if (!operationResultObserver.isDisposed()) {
-            operation.run(semaphore)
-                    .subscribeOn(subscribeScheduler)
-                    .unsubscribeOn(subscribeScheduler)
-                    .subscribe(new Observer<T>() {
-                        @Override
-                        public void onSubscribe(Disposable disposable) {
-                            /*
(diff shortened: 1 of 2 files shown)


dariuszseweryn:
@bleeding182 Did you confirmed that this change indeed helps?

bleeding182:
@DariuszAniszewski Hey there!

Yes, been debugging today, can't reproduce it with this fix.

dariuszseweryn:
It's @dariuszseweryn ;)

Great to hear that. I think that we should add a test for `FIFORunnableEntry` to prevent such a situation in the future.

The test should be straightforward to write and just verify that the `semaphore` is either released or passed to an operation depending on `ObservableEmitter.isDisposed()` result.

P.S. Maybe it would be a bit clearer if the code be something like:
```
if (observableEmitter.isDisposed()) {
  RxBleLog...
  semaphore.release();
  return;
}
operation.run(semaphore)...
```

rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/serialization/FIFORunnableEntry.java
```diff
@@ -35,44 +35,46 @@ public int compareTo(@NonNull FIFORunnableEntry other) {
     }
 
     public void run(QueueSemaphore semaphore, Scheduler subscribeScheduler) {
+
+        if (operationResultObserver.isDisposed()) {
+            RxBleLog.d("Operation was about to be run but the observer was already disposed: " + operation);
```
uKL: How about?

The operation was about to be run but the observer had been already disposed

PR #430, opened 2018-05-18
Fixes crash due to `Receiver not registered`

bleeding182:
Would happen on a 6P when toggling bluetooth on and off. This will make sure that we do always register before we unregister.


dariuszseweryn:
@bleeding182 I wonder if it could be possible to catch this use-case with a test... Have you checked perhaps?

bleeding182:
@DariuszAniszewski No, I did not look into writing any tests

The exception seems to be thrown when downstream is already disposed, hence it would immediately call unregister before we even registered.

By moving the unregister callback to the end this can't happen since we will always have registered by then

PR #435, opened 2018-05-26
Fix broken samples link

FireZenk:


dariuszseweryn:
Good spot! Thank you very much 👍

PR #478, opened 2018-09-14
Added notification quick setup mode

dariuszseweryn:
Some BLE peripherals are starting to notify right after the Command Characteristic Config Descriptor is written with ENABLE_NOTIFICATION value. In this case the DEFAULT notification setup mode could miss the first notifications before the downstream would subscribe. QUICK_SETUP mitigates this problem by postponing the write of Command Characteristic Config Descriptor until at least one observer has subscribed.

Parent `Observable<Observable<byte[]>>` emits an error if the scheduled CCC descriptor write fails and emitted child `Observable<byte[]>` completes. 

Connected to #195

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/connection/NotificationAndIndicationManager.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/connection/NotificationAndIndicationManager.java
@@ -1,6 +1,7 @@
 package com.polidea.rxandroidble2.internal.connection;
 
 
+import android.annotation.SuppressLint;
 import android.bluetooth.BluetoothGatt;
 import android.bluetooth.BluetoothGattCharacteristic;
 import android.bluetooth.BluetoothGattDescriptor;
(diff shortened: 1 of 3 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/connection/NotificationAndIndicationManager.java
```diff
@@ -138,17 +141,54 @@ public void run() {
...
+            @Override
+            public ObservableSource<Observable<byte[]>> apply(final Observable<Observable<byte[]>> upstream) {
+                switch (mode) {
+
+                    case COMPAT:
+                        return upstream;
+                    case QUICK_SETUP:
+                        final Completable publishedWriteCCCDesc = writeClientCharacteristicConfig(characteristic, descriptorWriter, value)
+                                .toObservable()
+                                .cache()
```
uKL: Is it safe? If you cache you give up the possibility of disposing of the observable.

dariuszseweryn: A good question. Writing CCC descriptor is a one time operation so it should end quickly. This question inspired me to write another test to check if writing the CCC will not happen after the parent `Observable<Observable<byte[]>>` is disposed

PR #501, opened 2018-10-28
Sample app improvements (runtime permissions, disposable handling, refactoring)

lukaszkalnik:
Add a runtime ACCESS_COARSE_LOCATION permission check and request for Android 6 devices to the sample app.
Also extract ble scan error handling to a separate helper class and extract error message strings to `strings.xml`.
Add disposable handling by disposing subscriptions in `onPause()`. Add some missing annotations.


lukaszkalnik:
I have entered my first and last name while signing the CLA and it seems to have been misinterpreted as a second contributor (`Lukasz Kalnik`). Could you please remove it? Thanks.

sample/src/main/java/com/polidea/rxandroidble2/sample/util/LocationPermission.java
```diff
@@ -0,0 +1,44 @@
...
+import android.support.v4.app.ActivityCompat;
+import android.support.v4.content.ContextCompat;
+
+public class LocationPermission {
+
+    private LocationPermission() {
+        // Utility class
+    }
+
+    private static int REQUEST_PERMISSION_COARSE_LOCATION = 9358;
```
dariuszseweryn:
```suggestion
    private static final int REQUEST_PERMISSION_COARSE_LOCATION = 9358;
```

sample/src/main/java/com/polidea/rxandroidble2/sample/util/ScanExceptionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+        // Special case, as there might or might not be a retry date suggestion
+        if (reason == BleScanException.UNDOCUMENTED_SCAN_THROTTLE) {
+            text = getUndocumentedScanThrottleErrorMessage(context, exception.getRetryDateSuggestion());
+        } else {
+            // Handle all other possible errors
+            final Integer resId = errorMessages.get(reason);
+            if (resId != null) {
+                text = context.getString(resId);
+            } else {
+                // unknown error - return default message
```
dariuszseweryn: An additional warning could be useful here as well I suppose

dariuszseweryn:
I cannot remove`Lukasz Kalnik` from contributors as one of your commits (`1a92c6e17d14751d60cfe1fc1ccf8972ff164f69`) was made with that user who cannot be attributed to your github account. To fix this situation you would need to change this commit's author to the proper one (`lukaszkalnik`)

lukaszkalnik:
Actually this commit doesn't change anything in perspective of the whole PR (just changes two Strings in two Activities, which I later moved anyway to strings.xml in another commit).
Can you remove it from history?

dariuszseweryn:
@lukaszkalnik Sorry for the delay — I was on a conference and short holiday. Actually you may and should use a force push to get rid of the commit by `Lukasz Kalnik` from this PR (afterall this PR is from your fork and you do have full control over it which I probably have not). Alternatively you may close this PR and create another one with different commits in it

lukaszkalnik:
I rebased and removed the commit (and force pushed) but the offending commit is still there... There is now just one merge remote commit on top.
Also when I created new PR from the same branch, the commit was still in the history, although I don't see it locally anymore.
Probably the best would be to just create a new branch from cherry-picked commits.

dariuszseweryn:
I have checkout your repo and I see this commit whose author is `lukasz.kalnik@grandcentrix.net` where the rest of your commits are from `lukasz.kalnik@gmail.com` so the history of your commits does not seem to be overwritten with via `push -f`. Quickly searched google and there seem to be [a fairly straightforward instruction on how you can change the author](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit) — I hope this helps

lukaszkalnik:
Thank you for your suggestions. Seems like the rebase in Source Tree doesn't work so well. I did it from the command line now and it worked :)

sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
```diff
@@ -122,4 +128,12 @@ private void updateUI() {
         connectButton.setText(connected ? R.string.disconnect : R.string.connect);
         autoConnectToggleSwitch.setEnabled(!connected);
     }
+
+    @Override
+    protected void onPause() {
+        super.onPause();
+
+        triggerDisconnect();
+        compositeDisposable.dispose();
```
dariuszseweryn:
```suggestion
        compositeDisposable.clear();
```
Otherwise the next time the activity will resume nothing will work

sample/src/main/java/com/polidea/rxandroidble2/sample/example4_characteristic/CharacteristicOperationExampleActivity.java
```diff
@@ -203,4 +214,10 @@ private boolean hasProperty(BluetoothGattCharacteristic characteristic, int prop
     private byte[] getInputBytes() {
         return HexString.hexToBytes(writeInput.getText().toString());
     }
+
+    @Override
+    protected void onPause() {
+        super.onPause();
+        compositeDisposable.dispose();
```
dariuszseweryn:
```suggestion
        compositeDisposable.clear();
```

sample/src/main/java/com/polidea/rxandroidble2/sample/util/ScanExceptionHandler.java
dariuszseweryn:
```suggestion
                // unknown error - return default message
                Log.w("Scanning", String.format("No message found for reason=%d. Consider adding one.", reason));
```
I was actually thinking about giving a warning log that we got a `reason` that has no error message so it will be clear what to fix

sample/src/main/java/com/polidea/rxandroidble2/sample/util/ScanExceptionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+            final Integer resId = ERROR_MESSAGES.get(reason);
+            if (resId != null) {
+                text = context.getString(resId);
+            } else {
+                // unknown error - return default message
+                text = context.getString(R.string.error_unknown_error);
+            }
+        }
+
+        Log.w("EXCEPTION", text, exception);
```
dariuszseweryn:
```suggestion
        Log.w("Scanning", text, exception);
```
Everyone will know that this is an exception because of a stacktrace ;)

PR #505, opened 2018-11-21
[Do not merge] Convert sample app to Kotlin

lukaszkalnik:
Please don't review this PR. The code here will be split in several smaller PRs which are easier to review. After all resulting partial PRs have been merged this PR can be closed.

~~This PR converts sample app to Kotlin.~~
~~I know it's a lot to review, but I thought it will be better to do in one go.~~
~~I think I will split this PR into PRs converting single classes to make it reviewable at all...~~

~~I have applied Android Studio automatic conversion to Kotlin. Afterwards I cleaned up the code and made it more idiomatic.~~

~~Please test the app thoroughly and check if I didn't mess up any of the logic. Thank you!~~


dariuszseweryn:
Thanks for the effort! I think it could be valuable to not to remove the `java` sample but simply add a `kotlin` one since we are in a kind of a transition period and people use both and users may not be accustomed to the new language yet. What do you think about the idea?

lukaszkalnik:
Closing as the whole kotlin sample app has been moved to the `sample-kotlin` module.

PR #512, opened 2018-11-29
[Sample] remove unnecessarily added Disposables from activities

lukaszkalnik:
Removes `Disposable` managing code unnecessarily added (by me) in previous PR with sample app improvements.
In most Activities the subscription has already been managed by binding it to specific events using `rxlifecycle`. This renders the additional `Disposable` handling code unnecessary and misleading to the reader.

[NOTE] duplicate suppressing of lint warnings in each case is for some reason required by Android Studio, otherwise it still complains about result not being used...

### DIFF
--- a/sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
+++ b/sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
@@ -1,5 +1,6 @@
 package com.polidea.rxandroidble2.sample.example2_connection;
 
+import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.os.Build;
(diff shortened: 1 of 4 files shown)


RobLewis:
I was just reading about Google's [Lifecycle-Aware Components](https://developer.android.com/topic/libraries/architecture/lifecycle#java) and wonder if they might be applicable. Especially with regard to disposing subscriptions.

lukaszkalnik:
@RobLewis theoretically you could use any MVP library, like [ThirtyInch](https://github.com/grandcentrix/ThirtyInch). Although like @dariuszseweryn wrote it's probably better to keep the examples simple and not mix too many concepts.
@dariuszseweryn I will remove `RxLifecycle` then 👍

lukaszkalnik:
I will just close this PR then and open another one.

PR #518, opened 2018-12-20
[fix] bump support library version to 28.0.0

lukaszkalnik:
`compileSdkVersion` in all modules is `28`, yet the support lib version was `27.1.0`.
This PR updates support lib version to `28.0.0`.


dariuszseweryn:
Thanks! The `develop` branch is already using `androidx` so hopefully it will get easier to maintain

PR #519, opened 2018-12-20
[sample] disposable handling

lukaszkalnik:
Removes `rxlifecycle` based Disposable handling, leaving only explicit handling in `onPause()`/`onDestroy()`.

Also adds missing `SDK` version check for background scanning calls.

### DIFF
--- a/sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
+++ b/sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
@@ -4,6 +4,7 @@ import android.annotation.TargetApi;
 import android.os.Build;
 import android.os.Bundle;
 import android.support.design.widget.Snackbar;
+import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.SwitchCompat;
 import android.widget.Button;
 import android.widget.EditText;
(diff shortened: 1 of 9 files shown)


sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
```diff
@@ -40,15 +36,15 @@
...
-    private final CompositeDisposable compositeDisposable = new CompositeDisposable();
+    private Disposable mtuDisposable;
+    private Disposable connectionStateDisposable;
 
     @OnClick(R.id.connect_toggle)
     public void onConnectToggleClick() {
         if (isConnected()) {
             triggerDisconnect();
         } else {
             connectionDisposable = bleDevice.establishConnection(autoConnectToggleSwitch.isChecked())
```
dariuszseweryn: This one does not seem to be properly disposed

sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
```diff
@@ -58,15 +54,12 @@ public void onConnectToggleClick() {
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     @OnClick(R.id.set_mtu)
     public void onSetMtu() {
-        final Disposable disposable = bleDevice.establishConnection(false)
+        mtuDisposable = bleDevice.establishConnection(false)
```
dariuszseweryn: It is possible that the `mtuDisposable` will get overwritten if user would tap the button more than once

sample/src/main/java/com/polidea/rxandroidble2/sample/example3_discovery/ServiceDiscoveryExampleActivity.java
```diff
@@ -39,7 +37,6 @@ public void onConnectToggleClick() {
         connectionDisposable = bleDevice.establishConnection(false)
```
dariuszseweryn: May get overwritten due to user clicks

sample/src/main/java/com/polidea/rxandroidble2/sample/example2_connection/ConnectionExampleActivity.java
lukaszkalnik: This one does get disposed by calling `triggerDisconnect()` in `onPause()`.

dariuszseweryn:
🚀

PR #531, opened 2019-01-16
[sample-kotlin] Gradle Kotlin DSL

lukaszkalnik:
Migrates the `sample-kotlin` module's `build.gradle.kts` to Gradle Kotlin DSL.

This PR builds on top of #527 , please merge it afterwards.


dariuszseweryn:
I see no particular advantages in updating gradle build script to kotlin — what is the motivation?

lukaszkalnik:
It was just an excercise for me. You're right, the build script is not that complex that it would give an advantage to the project.
And it seems to have caused different other inconveniences, so I think we could skip it.

You can close it :+1:

Issue #188, opened 2017-05-04
Direct connection is sometimes timeouted for 20s


misevcik:
### Summary
I've got following connection workflow:
1.) Start application
2.) Scan
3.) If device in range, stop scan. Direct connect to device

Sometimes (1 out if 10-20 attempts) connection is timeouted for 20s. Device I'm connecting to is in range and visible by scan. 
I got following error:
```
05:16:31.276 17413-17413/rxble E/BleClient: onConnectionFailure BleGattException{macAddress=00:16:A4:11:52:F1, status=133 (0x85 -> https://android.googlesource.com/platform/external/bluetooth/bluedroid/+/android-5.1.0_r1/stack/include/gatt_api.h), bleGattOperationType=BleGattOperation{description='CONNECTION_STATE'}}
```

Next connection attempt is always successful. 

#### Preconditions
Unknown
#### Steps to reproduce actual result
1.Unknown
#### Actual result
Sometimes connection is time outed.
#### Expected result
Connection is always successful.

###Log:
```
05:16:10.828 17413-17413/rxble I/MAIN: START
05:16:10.854 17413-17413/rxble D/BleScanner: Scan :true
05:16:10.856 17413-17413/rxble D/RxBle#Radio:   QUEUED RxBleRadioOperationScan(175046017)
05:16:10.857 17413-17626/rxble D/RxBle#Radio:  STARTED RxBleRadioOperationScan(175046017)
05:16:10.858 17413-17413/rxble I/FragmentContainer: CREATED
05:16:10.859 17413-17413/rxble D/ActivityThread: add activity client record, r= ActivityRecord{e168603 token=android.os.BinderProxy@a8a46b2 {rxble/rxble.MainActivity}} token= android.os.BinderProxy@a8a46b2
05:16:10.859 17413-17413/rxble I/MAIN: onResume
05:16:10.863 17413-17413/rxble I/MainActivity: onPause
05:16:10.871 17413-17413/rxble D/BluetoothAdapter: startLeScan(): null
05:16:10.871 17413-17413/rxble I/BluetoothAdapter: getBluetoothLeScanner
05:16:10.872 17413-17413/rxble I/BluetoothLeScanner: startScan 3 is called
05:16:10.874 17413-17413/rxble D/BluetoothAdapter: STATE_ON
05:16:10.881 17413-17425/rxble D/BluetoothLeScanner: onClientRegistered() - status=0 clientIf=5
[...]

PR #563, opened 2019-04-24
Updated RxAndroid dependency to remove Jettifier

### LINKED ISSUE #562: ERROR: [TAG] Failed to resolve variable '${animal.sniffer.version}'
### Disclaimer
Before you will post a new issue to the RxAndroidBle library consider if you are filling a bug/feature request or a general question 
about:
* Usage of RxJava
* Usage of Bluetooth Low Energy
* Usage of RxAndroidBle
* Weird behaviour of the peripheral you are using
* Other not directly related to a new feature request or a bug

If any of the above points seems to match your use case — consider using Google or creating a question on www.stackoverflow.com with 
a tag `rxandroidble` where it will be easier to access by other people with similar questions. Issues that are not bugs or feature requests
will be closed.

### Library version
`X.X.X`

<br />1.
<br />2.
<br />3.

dariuszseweryn:
Closes #562

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,4 +18,4 @@ POM_DEVELOPER_ID2=uKL
 POM_DEVELOPER_NAME2=Paweł Urban
 
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -22,7 +22,7 @@ ext {
             robolectric           : 'org.robolectric:robolectric:3.8',
             /* Sample apps-only dependencies */
             rxjava_replayingshare : 'com.jakewharton.rx2:replaying-share:2.1.1',
-            rxjava_binding        : 'com.jakewharton.rxbinding2:rxbinding:2.2.0',
+            rxjava_binding        : 'com.jakewharton.rxbinding3:rxbinding:3.0.0-alpha2',
             appcompat             : 'androidx.appcompat:appcompat:1.0.2',
             material_design       : 'com.google.android.material:material:1.0.0'
     ]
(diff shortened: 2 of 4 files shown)


gradle.properties
```diff
@@ -18,4 +18,4 @@ POM_DEVELOPER_ID2=uKL
 POM_DEVELOPER_NAME2=Paweł Urban
 
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
```
piotrek1543: I would keep it as `true` as long as Dagger2 does not support AndroidX :-)

dariuszseweryn: Hmmm... That's a good call — our shadowed version of Dagger2 may have not been properly checked. Isn't AndroidX supported in Dagger2 since `2.20`?

piotrek1543: Well, hard to say. The latest version of `dagger2` is as I see `2.22.1` so I would update it to that version and just check it, but there is no official info about full androidx support: https://github.com/google/dagger/issues/1188. 
Ron Shapiro in [#188](https://github.com/google/dagger/issues/1188) suggest using jetifier by default.

piotrek1543: I remember that few month ago I had issues with `2.16` version and setting `enableJetifier ` to `true` fixed the issue. I do not have experience with later versions as nowadays I am working on projects with `koin`.

dariuszseweryn: I have just checked with a fresh project and `1.9.2-SNAPSHOT` version:
```
./gradlew -Pandroid.enableJetifier=false canIDropJetifier

> Task :app:canIDropJetifier
========================================
Project app
========================================

No dependencies on old artifacts! Safe to drop Jetifier.
```
It compiled and run successfully. I suppose it is because the library does not use any support package classes and Dagger does not need to generate any code for them. No usage of those classes in generated code => no need for Jettifier.

dariuszseweryn: Anyway — thank you for double checking me 🥇

PR #625, opened 2019-09-20
Fixed typo in BleCharacteristicNotFoundException

dariuszseweryn:
This includes deprecating an accessor and creating a new one.

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/exceptions/BleCharacteristicNotFoundException.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/exceptions/BleCharacteristicNotFoundException.java
@@ -9,14 +9,30 @@ import java.util.UUID;
  */
 public class BleCharacteristicNotFoundException extends BleException {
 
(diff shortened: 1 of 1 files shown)


rxandroidble/src/main/java/com/polidea/rxandroidble2/exceptions/BleCharacteristicNotFoundException.java
```diff
@@ -9,14 +9,30 @@
...
     }
 
+    /**
+     * Returns characteristic UUID that has not been found
+     * @deprecated Use {@link #getCharacteristicUUID()}
+     *
+     * @return the UUID
+     */
+    @Deprecated
     public UUID getCharactersisticUUID() {
```
dar-trifork: missed one

dariuszseweryn: It is now deprecated. I cannot delete it until `2.0.0` as it would be a breaking change. It is advised to use getter without the typo.

Issue #668, opened 2020-02-25
Pass raw scanResult object so things like isConnectable can be called


juliansteenbakker:
**Is your feature request related to a problem? Please describe.**
Right now its not possible to access the isConnectable variable since its not getting parsed.

**Describe the solution you'd like**
Pass the raw scanResult object so we can access it.

**Describe alternatives you've considered**
Only pass isConnectable. However, other things like getDataStatus and getAdvertisementSid will not be accessible.

**Additional context**
[I already made a ticket in the react-native-ble-plx repository, but i found it that the issue originated here. ](https://github.com/Polidea/react-native-ble-plx/issues/619)

PR #669, opened 2020-02-25
Added raw scanResult object to custom scanResult 

juliansteenbakker:
so we can access things like isConnectable later on.
See issue #668


dariuszseweryn:
Hello @juliansteenbakker 
Sorry for such a delay. Thanks for your contribution. Unfortunately I do not think that it is a good way to go. It would be better to backport the functionality to legacy scan results as well. With the current approach we would get back to fragmentation problem.

juliansteenbakker:
> Hello @juliansteenbakker
> Sorry for such a delay. Thanks for your contribution. Unfortunately I do not think that it is a good way to go. It would be better to backport the functionality to legacy scan results as well. With the current approach we would get back to fragmentation problem.

I agree with what you are saying, but the scanResult object is added in API 21. Do you have anything in mind on how we can backport this to older api's?

dariuszseweryn:
Basically the same way it is implemented on Android side. [`ScanResult`](https://developer.android.com/reference/android/bluetooth/le/ScanResult) has a legacy constructor and shows legacy compatible behaviour.

PR #689, opened 2020-05-15
Add isConnectable to ScanResult

BeBetterBee:
Added isConnectable to scanResult, partial solution for issue #668

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/scan/RxBleInternalScanResult.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/scan/RxBleInternalScanResult.java
@@ -11,14 +11,16 @@ public class RxBleInternalScanResult {
     private final long timestampNanos;
     private final ScanRecord scanRecord;
     private final ScanCallbackType scanCallbackType;
+    private final Boolean isConnectable;
 
     public RxBleInternalScanResult(BluetoothDevice bluetoothDevice, int rssi, long timestampNanos, ScanRecord scanRecord,
-                                   ScanCallbackType scanCallbackType) {
+                                   ScanCallbackType scanCallbackType, Boolean isConnectable) {
         this.bluetoothDevice = bluetoothDevice;
         this.rssi = rssi;
         this.timestampNanos = timestampNanos;
         this.scanRecord = scanRecord;
         this.scanCallbackType = scanCallbackType;
+        this.isConnectable = isConnectable;
     }
 
     public BluetoothDevice getBluetoothDevice() {
@@ -39,5 +41,9 @@ public class RxBleInternalScanResult {
 
     public ScanCallbackType getScanCallbackType() {
         return scanCallbackType;
+    }
+
+    public Boolean isConnectable() {
+        return isConnectable;
     }
 }
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/scan/InternalScanResultCreator.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/scan/InternalScanResultCreator.java
@@ -9,6 +9,8 @@ import static com.polidea.rxandroidble2.scan.ScanCallbackType.CALLBACK_TYPE_UNKN
 import android.bluetooth.BluetoothDevice;
 import android.bluetooth.le.ScanResult;
 import android.bluetooth.le.ScanSettings;
+import android.os.Build;
+
 import androidx.annotation.RequiresApi;
 import androidx.annotation.RestrictTo;
 import com.polidea.rxandroidble2.ClientScope;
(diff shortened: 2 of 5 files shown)


marsounjan:
Hello @dariuszseweryn, what do you think about this PR? )) We'd appreciate it being processed since we need isConnectable in our project
thanks! :)

dariuszseweryn:
CLA isn't signed so I could not use it. Additionally this does not add functionality that would work for older devices  – which is what this library wanted to achieve — to reduce fragmentation as it is possible to add this functionality in a way that would work for older APIs.

BeBetterBee:
Sorry for not signing CLA, it is signed now. I am not sure if there is possibility to avoid fragmentation for isConnectable, it was added in API 26. We would like to use this feature, please consider adding it or propose other solution, thanks :)

dariuszseweryn:
You can see in [Android OS code](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/bluetooth/le/ScanResult.java;l=29?q=ScanResult&sq=) that the `ScanResult` has two constructors. One is legacy, one has `isConnectable` data. It can be implemented in the same way here.

marsounjan:
@dariuszseweryn I don't think it's as easy as you think. Before API 26 you just can't access the data you need for that :)

I think this answer on SO explains it pretty well - https://stackoverflow.com/a/28469205. Please check it out.

dariuszseweryn:
That's correct — yet, you can look on both constructors:
```
    /**
     * Constructs a new ScanResult.
     *
     * @param device Remote Bluetooth device found.
     * @param scanRecord Scan record including both advertising data and scan response data.
     * @param rssi Received signal strength.
     * @param timestampNanos Timestamp at which the scan result was observed.
     * @deprecated use {@link #ScanResult(BluetoothDevice, int, int, int, int, int, int, int,
     * ScanRecord, long)}
     */
    @Deprecated
    public ScanResult(BluetoothDevice device, ScanRecord scanRecord, int rssi,
            long timestampNanos) {
        mDevice = device;
        mScanRecord = scanRecord;
        mRssi = rssi;
        mTimestampNanos = timestampNanos;
        mEventType = (DATA_COMPLETE << 5) | ET_LEGACY_MASK | ET_CONNECTABLE_MASK;
        mPrimaryPhy = BluetoothDevice.PHY_LE_1M;
        mSecondaryPhy = PHY_UNUSED;
        mAdvertisingSid = SID_NOT_PRESENT;
        mTxPower = 127;
        mPeriodicAdvertisingInterval = 0;
    }

    /**
     * Constructs a new ScanResult.
     *
     * @param device Remote Bluetooth device found.
     * @param eventType Event type.
     * @param primaryPhy Primary advertising phy.
     * @param secondaryPhy Secondary advertising phy.
     * @param advertisingSid Advertising set ID.
     * @param txPower Transmit power.
     * @param rssi Received signal strength.
     * @param periodicAdvertisingInterval Periodic advertising interval.
     * @param scanRecord Scan record including both advertising data and scan response data.
     * @param timestampNanos Timestamp at which the scan result was observed.
     */
    public ScanResult(BluetoothDevice device, int eventType, int primaryPhy, int secondaryPhy,
            int advertisingSid, int txPower, int rssi, int periodicAdvertisingInterval,
            ScanRecord scanRecord, long timestampNanos) {
        mDevice = device;
        mEventType = eventType;
        mPrimaryPhy = primaryPhy;
        mSecondaryPhy = secondaryPhy;
        mAdvertisingSid = advertisingSid;
        mTxPower = txPower;
        mRssi = rssi;
        mPeriodicAdvertisingInterval = periodicAdvertisingInterval;
        mScanRecord = scanRecord;
        mTimestampNanos = timestampNanos;
    }
```
One has connectability info from `eventType` and the legacy one has a default `eventType`, yet both will return something when `.isConnectable()` is called. They will differ in `.isLegacy()` result.

marsounjan:
I don't get your point here. 
The legacy constructor makes all legacy results to return true on isConnectable() even when it's not true. So the developer must call isLegacy() first, just to check that isConnectable() returns relevant result...

I don't see a single reason why this should be better behaviour than @BeBetterBee proposed. Returning null in the case when we do not have the info about the connectability seems to me like the best way how add this functionality - even for older APIs.

dariuszseweryn:
Try to get info if `.isConnected()` gives relevant result from the proposed solution.

marsounjan:
@dariuszseweryn we have added check for `isLegacy()` on API26+ before checking `isConnectable()`. This way you always now that the isConnectable gives you the right result. 
 see it here (https://github.com/Polidea/RxAndroidBle/pull/689/commits/207d5d20394d50f5d13a4c7e8557fbfec8efe978)

Is that what you meant? :)

marsounjan:
@dariuszseweryn ^^ ?

dariuszseweryn:
That's definitely a better approach. I will think about best possible implementation and get back to you. (lately I do not have quite enough time unfortunately)

marsounjan:
@dariuszseweryn any update on this? :) ^^^

marsounjan:
@dariuszseweryn still no update here?)

Taym95:
Hi, @dariuszseweryn thanks for the great work, any possibility to merge this one and have isConnectable in ScanResult?

rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/scan/RxBleInternalScanResult.java
```diff
@@ -11,14 +11,16 @@
     private final long timestampNanos;
     private final ScanRecord scanRecord;
     private final ScanCallbackType scanCallbackType;
+    private final Boolean isConnectable;
 
     public RxBleInternalScanResult(BluetoothDevice bluetoothDevice, int rssi, long timestampNanos, ScanRecord scanRecord,
-                                   ScanCallbackType scanCallbackType) {
+                                   ScanCallbackType scanCallbackType, Boolean isConnectable) {
```
DougSig, line 17: I'm new here and am browsing but thought I'd offer a suggestion. To be more explicit, you could give it an enum instead of a nullable boolean like the following (pardon the kotlin):
```
enum class IsConnectableStatus { LEGACY_UNKNOWN, CONNECTABLE, NOT_CONNECTABLE }
```
Otherwise i'm not sure it's clear that `null` means pre-26.

RobLewis: Seems like a good solution.

jobfeikens:
Can we get an update on this? In my opinion a nullable boolean makes the most sense as that would indicate that we don't know whether the device is connectable

nrbrook:
I decided to fork this repository due to lack of activity. I have merged this PR (and others) and released as io.nrbtech.rxandroidble:rxandroidble:2.0.0 on Maven Central, repository here: https://github.com/NRB-Tech/RxAndroidBle

dariuszseweryn:
I ended using #781 
Sorry that it took so long. I will create a release Today probably

dariuszseweryn:
Currently available in 1.14.0-SNAPSHOT

PR #708, opened 2020-07-31
Update ScanOperationApi21.java - Fixing the memory leak from ScanOperationApi21 emitter

seotrader:
Fixing the memory leak from ScanOperationApi21 emitter
Fixing the issue: https://github.com/Polidea/RxAndroidBle/issues/607

### DIFF
--- a/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/operations/ScanOperationApi21.java
+++ b/rxandroidble/src/main/java/com/polidea/rxandroidble2/internal/operations/ScanOperationApi21.java
@@ -38,6 +38,8 @@ public class ScanOperationApi21 extends ScanOperation<RxBleInternalScanResult, S
     final EmulatedScanFilterMatcher emulatedScanFilterMatcher;
     @Nullable
     private final ScanFilter[] scanFilters;
+    @Nullable
+    private ObservableEmitter<RxBleInternalScanResult> scanEmitter;
 
 
     public ScanOperationApi21(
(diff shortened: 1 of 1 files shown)


dariuszseweryn:
Hello!
I am trying to get back to this issue with the author of Leak Canary as I have some doubts.
There is one flaw in your PR — it is prone to race condition when the callback is being nullified on a different thread.

seotrader:
Hi @dariuszseweryn , any updates from Leak Canary?

Regarding to the race condition on the callback. Do you have a preferred way for me to fix it?

Thanks

dariuszseweryn:
No updates yet, I am fully occupied :/
As for the preferred way — a simple fix would be to have a local reference in each function before checking if the recipient is not null.

b055man:
@dariuszseweryn I'm observing a problem with the library after prolonged used, namely after several hours spent on reconnect attempts (no scanning, but thousands of connect attempts) - it looks like afterwards flutter_ble_lib is not operational anymore and any attempt to initate scanning fails instantly as the stream is closed (onDone is called instantly).
Looking at Android Studio noticed this: 
image
This is just a heads up - I will try to provide more data and observations. For now, just a question if you've seen something smilar before?

seotrader:
@dariuszseweryn made another PR with the fix you wanted

I don't see any leak with LeakCanary after calling scanBleDevices() and closing the activity

image

dariuszseweryn:
@b055man Have you reported this issue on Flutter lib repo? I have not seen anything like that yet.
@seotrader Merging. 👍  Thank you

b055man:
@dariuszseweryn Sorry for not updating this. The root cause for this problem has been identified already - I've an issue here: https://github.com/Polidea/FlutterBleLib/issues/528 The original cause has been resolved in the RxAndroidBle RxJava2-based version of the lib but FlutterBleLIb/MultiplatformBleAdapter was (still is, to be precise) based on the earlier version.

PR #762, opened 2021-07-30
WIP: Support API 31 (Android 12) new Bluetooth permissions

### LINKED ISSUE #752: Support Android 12 Bluetooth Permissions
With Android 12 comes two new bluetooth permissions that will replace Location Permission:

* Bluetooth Scan
* Bluetooth Connect

The library needs to be updated to reflect these states.  In addition, Bluetooth Connect is only
required if you're going to open a gatt connection to the the device, so it shouldn't block the
operation of the library.  Currently the RxBleClient.state is a progressive check that ends up with an isReadyForScanning
when all permissions and services have turned on.

I'd recommend that this be broken out into two states.  Once for scanning and one for connecting.

https://developer.android.com/about/versions/12/features/bluetooth-permissions

dariuszseweryn:
API 31 introduces new Bluetooth permissions required for BLE operations. In the context of RxAndroidBle the important ones are:
- `BLUETOOTH_SCAN` — replaces previously needed ACCESS_FINE_LOCATION
- `BLUETOOTH_CONNECT` — a new concept

Considerations:
The library's AndroidManifest.xml does include `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION` since API 23. Technically these could be restricted to `maxSdkVersion="30"` but it could be a breaking change for apps that require location access otherwise. There is a new example in the Readme if the user does not want to require location access for other reasons than scanning BLE peripherals.

TODO:
- [x] Check if having `BLUETOOTH_SCAN` makes Location Services requirement obsolete
Result: If the user sets flag `neverForLocation` then Location Services are not needed. Otherwise they still are needed to get scan results.
- [x] Adjust the code to conform to new requirements for Location Services
- [x] ~Add check for new `BLUETOOTH_CONNECT` permission?~ Expected to be a thrown exception anyway
- [x] ~Design new `RxBleClient#State` API for convenient usage with the new permission model~ Extracted to #764 

Resolves #752

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,11 @@ Runtime permissions required for running a BLE scan:
 | 18 | 22 | (No runtime permissions needed) |
 | 23 | 28 | One of below:<br>- `android.permission.ACCESS_COARSE_LOCATION`<br>- `android.permission.ACCESS_FINE_LOCATION` |
 | 29 | 30 | - `android.permission.ACCESS_FINE_LOCATION`<br>- `android.permission.ACCESS_BACKGROUND_LOCATION`\* |
-| 31 | current | - `android.permission.BLUETOOTH_SCAN`\*\* |
+| 31 | current | - `android.permission.BLUETOOTH_SCAN`\*\*<br>- `android.permission.ACCESS_FINE_LOCATION`\*\*\* |
 
 \* Needed if [scan is performed in background](https://developer.android.com/about/versions/10/privacy/changes#app-access-device-location)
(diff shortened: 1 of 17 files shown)


b055man:
Hi @dariuszseweryn,

thanks for working on this!
Is there any ETA for publishing this change? Better yet, any chance for getting a new version of multiplatform_ble_adapter + flutter_ble_lib once that change is merged.. ? :)

dariuszseweryn:
Hi, I was on long vacations. I am trying to maintain only the RxAndroidBle. Other mentioned libs were transferred to a different entity...

PR #782, opened 2022-04-21
Compliant with Android 12 scan device permission requirements

ZacharyTech:
- Solve the sample scan device on Android 12
- Add scan device unexpected exception log


ZacharyTech:
Submitting this PR is just to provide a solution for Android 12 scan failures.

